PR #48, opened 2017-11-09
Fix wrapping of left aligned text

### LINKED ISSUE #47: Wrapped text should not include space at start of new line
_originally posted by @CDDelta at  SixLabors/ImageSharp#375_

### Description

When wrapping text with spaces to a new line DrawText() starts the next line with a space.

### Steps to Reproduce

Draw text (eg. "TEST TEST TEST TEST") that includes spaces and enable wrapping.
Output and view it.
Here is an example I made.
output_512x512

CDDelta:
Fixes wrapping for left aligned text. Is there a need for new tests? If so how would I go about it?

Fixes #47

### DIFF
--- a/src/SixLabors.Fonts/TextLayout.cs
+++ b/src/SixLabors.Fonts/TextLayout.cs
@@ -83,9 +83,15 @@ namespace SixLabors.Fonts
 
                 char c = text[i];
 
-                if (char.IsWhiteSpace(c))
+                // if text is supposed to be left aligned, only wrap the text after the whitespace
+                if (options.HorizontalAlignment == HorizontalAlignment.Left)
                 {
-                    //find the index in the layout where we last enabled back tracking from
+                    if (i > 0 && char.IsWhiteSpace(text[i - 1]))
+                        lastWrappableLocation = layout.Count;
+                }
+                else if (char.IsWhiteSpace(c))
+                {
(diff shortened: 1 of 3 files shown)


src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -83,9 +83,15 @@ public ImmutableArray<GlyphLayout> GenerateLayout(string text, RendererOptions o
 
                 char c = text[i];
 
-                if (char.IsWhiteSpace(c))
+                // if text is supposed to be left aligned, only wrap the text after the whitespace
+                if (options.HorizontalAlignment == HorizontalAlignment.Left)
                 {
-                    //find the index in the layout where we last enabled back tracking from
+                    if (i > 0 && char.IsWhiteSpace(text[i - 1]))
+                        lastWrappableLocation = layout.Count;
```
tocsoft: curly braces please... our codestyle requires curly braces even on single statement ifs

CDDelta:
@tocsoft I was wondering whether or not it would be better to remove the whitespace from the end of the line to avoid problems with TextMeasurer. If you look at TextLayoutTests.MeasureTextWordWrapping, TextMeasurer returns 360 width which is longer than the expected max 350. Would you consider this unwanted behaviour?

CDDelta:
I have opened a PR #49 that I think handles this issue better.

PR #49, opened 2017-11-10
Remove white space from wrapped text end

### LINKED ISSUE #47: Wrapped text should not include space at start of new line

CDDelta:
Removes whitespace from the end of wrapped text lines. Looks neater when rendered with all different alignments and provides more accurate TextMeasurer results.

Fixes #47

### DIFF
--- a/src/SixLabors.Fonts/TextLayout.cs
+++ b/src/SixLabors.Fonts/TextLayout.cs
@@ -31,6 +31,9 @@ namespace SixLabors.Fonts
             float originX = 0;
             if (options.WrappingWidth > 0)
             {
+                // trim trailing white spaces from the text
+                text = text.TrimEnd(null);
+
                 maxWidth = options.WrappingWidth / options.DpiX;
 
                 switch (options.HorizontalAlignment)
--- a/src/SixLabors.Fonts/TextLayout.cs
+++ b/src/SixLabors.Fonts/TextLayout.cs
@@ -83,7 +83,7 @@ namespace SixLabors.Fonts
 
                 char c = text[i];
 
-                if (char.IsWhiteSpace(c))
+                if (options.WrappingWidth > 0 && char.IsWhiteSpace(c))
                 {
                     // keep a record of where to wrap text and ensure that no line starts with white space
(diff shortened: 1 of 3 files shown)


CDDelta:
Text such as "hello   world   hello   world" when wrapped might end up starting with white space, should this be expected behaviour or should I make sure that it does not happen?

tocsoft:
I feel we should make sure we drop all white space from the start of wrapped lines and end of lines.
_`_` represents a space_
thus 
```
_ _ _hello_ _ _world_ _ _hello_ _ _world_ _ _hello_ _ _
```

would wrap to 
### left aligned
``` 
_ _ _ hello_ _ _world
hello_ _ _world
hello
```
### right aligned
``` 
_ _ _ hello_ _ _world
      hello_ _ _world
                hello
```
### center aligned
``` 
_ _ _ hello_ _ _world
  hello_ _ _world    
       hello
```
but I feel we want to make sure we don't don't trim any trailing white space from text when not in text wrapping mode.

CDDelta:
Since white space trimming is done in the wrapping loop, there should be no effect on non-wrapped text.

tocsoft:
@CDDelta been having a play with this looking good. One small snag I can see is we are not ignoring the trailing white space when laying out the last line.

I've updated the 'DrawWithImageSharp' sample in master to include this output to help you visualise it.

alignmentwrapped - copy

CDDelta:
@tocsoft updated to trim white space. Anything else?

CDDelta:
@tocsoft first ever pull request :) Thank you for your great work on this library.

tocsoft:
your welcome.. we will get this fix released to nuget soon, we have some plans in the works for new beta of ImageSharp so it's likely to get included in that wave however it'll be available on our myget feed shortly immediately.

PR #54, opened 2018-01-03
Load fonts without case sensitivity on all platforms

### LINKED ISSUE #53: SystemFontCollection: File names/extensions are case sensitive under Linux
Problem:
---

When using SystemFontCollection, [On This Line](https://github.com/SixLabors/Fonts/blob/94d9db66c98ef843131ebb5d879475a5138580ee/src/SixLabors.Fonts/SystemFontCollection.cs#L41), it'll try to look for fonts using the *.ttf filter.

That will fail on Unix systems if the font file has a *.TTF extension (uppercase) instead, since its has a case sensitive file system.

Possible Solution:
---

Alternatively, replace that line with

```csharp
IEnumerable<string> files = 
    found.SelectMany(x => Directory.EnumerateFiles(x, "*.*", SearchOption.AllDirectories))
         .Where(x => Path.GetExtension(x).Equals(".ttf", StringComparison.OrdinalIgnoreCase));
```

So its more reliable on platforms other than windows.

Tested on:
---
* .NET Core 2.0.4-servicing-25915-01 @ Raspbian (armhf)
* .NET Core 2.0.0 @ Debian 9 (x64)

*edited to perma-link the source line reference*

CDDelta:
Implements the fix by @usagirei to load fonts without case sensitivity on all platforms, specifically Linux.

Fixes #53

### DIFF
--- a/src/SixLabors.Fonts/SystemFontCollection.cs
+++ b/src/SixLabors.Fonts/SystemFontCollection.cs
@@ -38,7 +38,10 @@ namespace SixLabors.Fonts
             string[] expanded = paths.Select(x => Environment.ExpandEnvironmentVariables(x)).ToArray();
             string[] found = expanded.Where(x => Directory.Exists(x)).ToArray();
 
-            IEnumerable<string> files = found.SelectMany(x => Directory.EnumerateFiles(x, "*.ttf", SearchOption.AllDirectories));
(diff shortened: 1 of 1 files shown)


tocsoft:
@CDDelta thanks for the PR, much appreciated.

PR #62, opened 2018-04-24
Extended TextMeasurer to be able to measure character bounds.

woutware:


src/SixLabors.Fonts/GlyphMetric.cs
```diff
@@ -0,0 +1,46 @@
...
+// Licensed under the Apache License, Version 2.0.
+
+using SixLabors.Primitives;
+
+namespace SixLabors.Fonts
+{
+    /// <summary>
+    /// Represents a <see cref="Glyph"/> metric.
+    /// </summary>
+    public struct GlyphMetric
```
JimBobSquarePants: If you mark the struct as `readonly` this will prevent the compiler creating defensive copies. We generally ensure all structs implement `IEquatable<T>` also to allow efficient comparison in the future.

woutware: Made struct readonly.

I don't see Equals ever being used for this struct though, as it concerns glyph layout info. For a low level structs like points and vectors I can see the use of having a type safe Equals method.

tocsoft:
@woutware thanks a lot 👍

PR #63, opened 2018-04-24
Added IFontInstance properties Ascender, Descender and LineGap.

woutware:


tocsoft:
thanks @woutware 👍 you will be able to access these changes from our myget feed until we next do a full release

PR #67, opened 2018-05-22
Fix vertical layout problems.

woutware:
There was a gap between the top of the line and the top of the glyphs of the first line that the vertical alignment didn't take into account. Also there was a minor bug in calculating the line height itself that possibly happened when there were glyphs of varying line heights in the line.


JimBobSquarePants:
Thanks for this! 👍 

Looks like we've got some failing tests that need to be updated to match the new output. 

Note to self, fix all the stylecop errors so the build output is less cluttered.

woutware:
I am going over this again right now, I might also need to take into account the font descender in addition to the ascender. Strange that the output looks ok though.

tocsoft:
I'm going to be doing a few spots of repo spring cleaning and should get an updated beta out this evening for Fonts

PR #72, opened 2018-06-20
Fix operators and hashcode

JimBobSquarePants:
`==` operator was incorrect and we were still calling `base.GetHashCode()`

### DIFF
--- a/src/SixLabors.Fonts/GlyphRendererParameters.cs
+++ b/src/SixLabors.Fonts/GlyphRendererParameters.cs
@@ -3,14 +3,12 @@
 
 using System;
 using System.Numerics;
-using System.Runtime.CompilerServices;
-using SixLabors.Primitives;
 
 namespace SixLabors.Fonts
 {
     /// <summary>
     /// The combined set of properties that uniquely identify the glyph that is to be rendered
-    /// at a particular sixe and dpi.
+    /// at a particular size and dpi.
     /// </summary>
     public readonly struct GlyphRendererParameters : IEquatable<GlyphRendererParameters>
     {
@@ -67,7 +65,7 @@ namespace SixLabors.Fonts
         /// True if the current left is equal to the <paramref name="right"/> parameter; otherwise, false.
         /// </returns>
         public static bool operator ==(GlyphRendererParameters left, GlyphRendererParameters right)
-            => !left.Equals(right);
+            => left.Equals(right);
 
         /// <summary>
         /// Compares two <see cref="GlyphRendererParameters"/> objects for inequality.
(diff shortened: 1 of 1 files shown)


src/SixLabors.Fonts/GlyphRendererParameters.cs
```diff
@@ -85,36 +83,27 @@ internal GlyphRendererParameters(GlyphInstance glyph, float pointSize, Vector2 d
...
 
         /// <inheritdoc/>
         public override int GetHashCode()
         {
             var hash = this.Font?.GetHashCode() ?? 0;
             hash = HashHelpers.Combine(hash, this.PointSize.GetHashCode());
             hash = HashHelpers.Combine(hash, this.DpiX.GetHashCode());
-            hash = HashHelpers.Combine(hash, this.DpiY.GetHashCode());
-            return base.GetHashCode();
+            return HashHelpers.Combine(hash, this.DpiY.GetHashCode());
```
tocsoft: we need to add the `FontStyle`  and `GlyphIndex` to match the equals for the correct hash code.

src/SixLabors.Fonts/GlyphRendererParameters.cs
```diff
@@ -85,36 +83,27 @@ internal GlyphRendererParameters(GlyphInstance glyph, float pointSize, Vector2 d
...
-            }
-
-            return false;
-        }
+        public override bool Equals(object obj) => obj is GlyphRendererParameters p && this.Equals(p);
 
         /// <inheritdoc/>
         public override int GetHashCode()
         {
             var hash = this.Font?.GetHashCode() ?? 0;
```
tocsoft: we should probably lowercase the font name when hash coding to match the `OrdinalIgnoreCase` matching in the equals.

src/SixLabors.Fonts/GlyphRendererParameters.cs
```diff
@@ -67,7 +65,7 @@ internal GlyphRendererParameters(GlyphInstance glyph, float pointSize, Vector2 d
         /// True if the current left is equal to the <paramref name="right"/> parameter; otherwise, false.
         /// </returns>
         public static bool operator ==(GlyphRendererParameters left, GlyphRendererParameters right)
-            => !left.Equals(right);
+            => left.Equals(right);
```
tocsoft: woops.. that would be my copy pasting haste

JimBobSquarePants:
Yeah, spotted it cos I'm trying to make a more conscious effort to understand the library so I can help out there more. Next will be to replace the custom big endian code with BinaryPrimitives.!

Issue #57, opened 2018-02-06
Feature Request: Expose font glyphs, geometry, etc


vpenades:
Although the main focus of SixLabors.Fonts is to provide font rendering features to ImageSharp, I don't see why it can't be used for much more than that.

So I would like to request to expose glyph information such as:
- Glyph Metrics
- Glyph Geometry

Glyph metrics are required in case you want to create a texture font atlas, as used by game engines, where each letter is rendered into the texture, and the corresponding glyph metadata is stored in a separate file, so the engine can use it to render text on its own.

Glyph geometry (both raw and triangulated) can be used for vector graphics, or rendering large form factor text in videogames.

tocsoft:
As a work around for you you can use the `SixLabors.Shapes.Text` package to render the text to Vectors which you can transform/interrogate as required.

woutware:
We need access to the font em size, ascent, descent, and line spacing, analogous to methods `int System.Drawing.FontFamily.GetEmHeight(System.Drawing.FontStyle)`, `int System.Drawing.FontFamily.GetCellAscent(System.Drawing.FontStyle)`, `int System.Drawing.FontFamily.GetCellDescent(System.Drawing.FontStyle)`, `int System.Drawing.FontFamily.GetLineSpacing(System.Drawing.FontStyle)`.

The use case is to do our own text layout based on these metrics to mimic AutoCAD's text rendering, which deviates slightly from the raw true type font metrics, but we use these metrics as a starting point, which is then slightly modified.

tocsoft:
This can/should be done by newing up a `Font` instance (as this represents a styled font)

we already expose [EmSize](https://github.com/SixLabors/Fonts/blob/16f933e014cd62f5e593132fba6268b710629551/src/SixLabors.Fonts/Font.cs#L117), and with only a single small addition of `public int LineHeight => this.instance.Value.LineHeight;` can expose the line height. 

we should just need to expand https://github.com/SixLabors/Fonts/blob/master/src/SixLabors.Fonts/FontInstance.cs and https://github.com/SixLabors/Fonts/blob/master/src/SixLabors.Fonts/IFontInstance.cs to expose the final 2 values and should be simple enough.

woutware:
https://github.com/SixLabors/Fonts/pull/63

JimBobSquarePants:
Fixed by #63 and #102

PR #77, opened 2018-06-23
Expose glyph size

Jjagg:
I'm working on a Font atlas implementation so I needed to get the glyph size. Hope it's alright to expose it like this :) 

Related to #57. I think it would be useful to expose some more glyph data through `Glyph` as mentioned in that issue (though this is all I needed for now).

### DIFF
--- a/src/SixLabors.Fonts/Font.cs
+++ b/src/SixLabors.Fonts/Font.cs
@@ -145,13 +145,13 @@ namespace SixLabors.Fonts
         internal IFontInstance FontInstance => this.instance.Value;
 
         /// <summary>
-        /// Gets the glyph.
+        /// Gets the glyph that corresponds to the given character.
         /// </summary>
-        /// <param name="codePoint">The code point of the character.</param>
+        /// <param name="character">The UTF-32 encoded character.</param>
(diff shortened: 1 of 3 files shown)


tocsoft:
Could you not have used the `TextMeasurer.TryMeasureCharacterBounds` method?
https://github.com/SixLabors/Fonts/blob/2629e4469be58308532dc89319bbbab67df4d0cd/src/SixLabors.Fonts/TextMeasurer.cs#L43

Jjagg:
I could have, but since I just need the size of individual glyphs the whole layout step is overkill.

tocsoft:
Thanks for your PR but i'm afraid i'm not currently comfortable accepting it, Sorry. 

#### My reasons:

We already have a public API that can provide you with the data you need without having to expose internal implementation details on to the public API. 

I'm trying to make the public API surface a lean as possible to reduce the number of potential breaking changes that will be required to support more/all font formats and variations.

You will find the layout portion of the code will be so minor in the grand scheme of rendering glyphs (rasterization is where the time will be spent) that I really wouldn't worry about it, text layout is highly unlikely to be your bottleneck.

Jjagg:
Okay, I understand, no worries!

PR #78, opened 2018-06-29
Return font description when installing font

woutware:


tocsoft:
@woutware all merged in, I fixed the failing build and tweeked some little bits but all in. thanks 👍

woutware:
Thanks for the quick merge!

PR #83, opened 2018-08-23
Improve CQ

iamcarbon:
This PR is pretty big.

Main changes:

- Factors out the EndianBitConverter with faster BinaryPrimitives
- Makes structs readonly when possible
- Passes these structures by readonly reference where possible
- Uses TryGetValue where possible
- Removes a ton of dead code
- Eliminates various allocations
- Tidies up everything


iamcarbon:
@tocsoft  -- this is ready for review / feedback.

Going to see how hard it will be to work in woff2 support after this.

PR #84, opened 2018-08-26
Improve CQ2

iamcarbon:
- Formats code using ternary operators and expressions
- Eliminate segment allocations
- Remove HashHelpers (exists in Core)


iamcarbon:
@tocsoft Ready for review. I think this covers all the nits.

src/SixLabors.Fonts/FontDescription.cs
```diff
@@ -90,16 +90,16 @@ public static FontDescription LoadDescription(Stream stream)
         /// </returns>
         internal static FontDescription LoadDescription(FontReader reader)
         {
-            HeadTable head = reader.GetTable<HeadTable>();
-            OS2Table os2 = reader.GetTable<OS2Table>();
-            NameTable nameTable = reader.GetTable<NameTable>();
-
-            return new FontDescription(nameTable, os2, head);
+            return new FontDescription(
```
tocsoft: the order we are loading these tables in was (vaguely) important... basically there is an recommended terminated table order. https://docs.microsoft.com/en-gb/typography/opentype/spec/recom#optimized-table-ordering

Thats why it was head, os2 and name tables in that order. Your change inverts the order which will cause us to generally have to just backwards in the stream more often.

We should revert this change... we should probably comment it as well so we don't forget in the future.

tocsoft:
thanks 👍

PR #91, opened 2018-11-10
Fix incorrect glyph index check

### LINKED ISSUE #90: cmap table doesn't support 32-bit characters yet.
```
ushort GetGlyphIndex(int codePoint)
{
            if (codePoint > short.MaxValue)
            {
                throw new NotImplementedException("cmap table doesn't support 32-bit characters yet.");
            }

            return this.cmap.GetGlyphId(codePoint);
 }
```
There are errors here,it should be `ushort.MaxValue`
```
if (codePoint > short.MaxValue)=>if (codePoint > ushort.MaxValue)
```

Jjagg:
Fixes #90 

Sorry, this was a pretty bad mess up... 😅

### DIFF
--- a/src/SixLabors.Fonts/FontInstance.cs
+++ b/src/SixLabors.Fonts/FontInstance.cs
@@ -86,7 +86,7 @@ namespace SixLabors.Fonts
 
         internal ushort GetGlyphIndex(int codePoint)
         {
-            if (codePoint > short.MaxValue)
+            if (codePoint > ushort.MaxValue)
             {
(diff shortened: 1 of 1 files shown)


tocsoft:
👍 thanks for the fix

rlightner:
I'm trying to print an icon from Font Awesome 5 woff file and the indexes seem to be off. Trying to print `f6ac` (wifi-slash) and it's printing the `f6ab` (wifi-2) character instead.

PR #95, opened 2018-12-12
Use HashCode.Combine

iamcarbon:


src/SixLabors.Fonts/GlyphRendererParameters.cs
```diff
@@ -100,12 +100,13 @@ public bool Equals(GlyphRendererParameters other)
...
         public override int GetHashCode()
         {
-            var hash = this.Font?.GetHashCode() ?? 0;
-            hash = HashHelpers.Combine(hash, this.PointSize.GetHashCode());
-            hash = HashHelpers.Combine(hash, this.GlyphIndex.GetHashCode());
-            hash = HashHelpers.Combine(hash, this.FontStyle.GetHashCode());
-            hash = HashHelpers.Combine(hash, this.DpiX.GetHashCode());
-            return HashHelpers.Combine(hash, this.DpiY.GetHashCode());
+            return HashCode.Combine(
+                this.Font,
```
dlemstra: Can `HashCode.Combine` handle null values?

iamcarbon: Yep! Combines with 0 if it's null.

Issue #65, opened 2018-05-17
Make FontInstance public


tzachshabtay:
Currently FontInstance is internal, meaning I can't call the LoatFont variants from my code and I'm forced to use FontCollection.
However FontCollection is not a useful abstraction for me: it first loads the font from the stream and then checks if the family name already exists.
I know better in my code and have my own caching to prevent loading the font from the stream in the first place (using file name or embedded resource id as the key) if it's already loaded so I don't need FontCollection and it just adds overhead with those dictionary checks and the lock: https://github.com/SixLabors/Fonts/blob/4ee628bad5c0206440bd6cc865ee3d6a3285c329/src/SixLabors.Fonts/FontCollection.cs#L105

Making FontInstance public would allow me to implement my own caching without those overheads.

JimBobSquarePants:
Fixed by #102

PR #102, opened 2019-01-05
Expose glyph data

### LINKED ISSUE #89: Glyph Functionality Questions
I have a few questions which should probably be turned into documentation and samples down the line as currently, they aren't very clear (at least to me - quite possible I missed it).

How would one go about getting the metrics of a single glyph within a font? The Font class' `GetGlyph` function is internal, and it returns an object which doesn't have them (which would be GlyphInstance, I guess?).

There is a sample showing how to render text, but how do you render a single glyph? I would like to render the glyphs to an atlas and then render that using something else.

feliwir:
I need this to rendering fonts myself with a GPU rasterizer. Since this library is not just about `rasterizing` fonts but also about `loading` them it seems useful to expose those informations. In the current version it's not possible to obtain any informations about glyphs apart from their dimensions.

Fixes #89 #65 #57

### DIFF
--- a/src/SixLabors.Fonts/Glyph.cs
+++ b/src/SixLabors.Fonts/Glyph.cs
@@ -20,7 +20,7 @@ namespace SixLabors.Fonts
         /// <value>
         /// The glyph instance.
         /// </value>
-        public GlyphInstance GlyphInstance => this.instance;
+        public GlyphInstance Instance => this.instance;
 
         internal Glyph(GlyphInstance instance, float pointSize)
         {
--- a/src/SixLabors.Fonts/Glyph.cs
+++ b/src/SixLabors.Fonts/Glyph.cs
@@ -9,10 +9,18 @@ namespace SixLabors.Fonts
     /// <summary>
     /// A glyph from a particular font face.
     /// </summary>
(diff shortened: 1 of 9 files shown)


src/SixLabors.Fonts/Glyph.cs
```diff
@@ -9,11 +9,19 @@ namespace SixLabors.Fonts
...
         private readonly GlyphInstance instance;
         private readonly float pointSize;
 
+        /// <summary>
+        /// Gets the glyph instance.
+        /// </summary>
+        /// <value>
+        /// The glyph instance.
+        /// </value>
+        public GlyphInstance GlyphInstance => this.instance;
```
iamcarbon: Glyph prefix is redundant. 

```
public GlyphInstance Instance => this.instance;
```

feliwir: Should `FontInstance` in Font also be renamed?

JimBobSquarePants: I would say so yes 👍

feliwir: Did the required change

PR #104, opened 2019-01-15
Fix MyGet link in readme

Jjagg:
The casing was off so the link didn't work.


tocsoft:
thanks for this fix 👍

PR #116, opened 2019-10-25
Remove dependency on SixLabors.Core

tocsoft:
Breaking public API Changes
- IGlyphRenderer change PointF to Vector2
- IGlyphRenderer change RectangleF to FontRectangle
- TextMeasurer change SizeF to FontRectangle
- TextMeasurer change RectangleF to FontRectangle

Once this one is in we can then update the Drawing/Shapes code to cope with the breaking changes.

Relates to SixLabors/ImageSharp#1002


src/SixLabors.Fonts/TextMeasurer.cs
```diff
@@ -22,7 +21,7 @@ public static class TextMeasurer
         /// <param name="text">The text.</param>
         /// <param name="options">The style.</param>
         /// <returns>The size of the text if it was to be rendered.</returns>
-        public static SizeF Measure(string text, RendererOptions options)
+        public static FontRectangle Measure(string text, RendererOptions options)
```
antonfirsov: What is the value of `rect.X` and `rect.Y` here? Are we sure this is better, than returning `Vector2` here?

tocsoft: Its actually the top left position based on the render... no problem with just setting X,Y to 0,0 either, couldn't decide when I made the change so I left it as the render position.

The benefit to using this instead of `Vectort2` is for anyone using `var` instead of the type directly then the returned type still has a `Width` and `Height` property which is a much more intuitive experience and thus source code compatible.

JimBobSquarePants: Man... I can't make up my mind about `FontRectangle`...

src/SixLabors.Fonts/SixLabors.Fonts.csproj
```diff
@@ -27,7 +27,8 @@
     <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
     <GenerateAssemblyInformationalVersionAttribute>false</GenerateAssemblyInformationalVersionAttribute>
     <DebugType Condition="$(codecov) != ''">full</DebugType>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>8.0</LangVersion>
+    <Nullable>enable</Nullable>
```
JimBobSquarePants: Any issues highlighted with this?

tocsoft:
Not all the usages of FontRect are size only we have APIs that are turning a rect in 2d space. IE top left plus height and width.

JimBobSquarePants:
>Not all the usages of FontRect are size only we have APIs that are turning a rect in 2d space. IE top left plus height and width.

Shouldn't they be `Vector4` then? I might be getting myself confused here...

tocsoft:
As a public API representing something with a height and width vector4 would be and awful experience. We couldn't even use intellisense to let people know that z and w mean height and width.

JimBobSquarePants:
That's the crux of my issue around converting `SizeF` to `Vector2`. 

This method is called `Size` but when consumed, once passed around it's going to  rely on good naming to convey any meaning.

https://github.com/SixLabors/Fonts/pull/116/files#diff-1d831cf1aa4bf91dc4230a74a6de57dcR25

tocsoft:
That's an internal class not part of the public API. This PR was about the external API not internals.

src/SixLabors.Fonts/SixLabors.Fonts.csproj
tocsoft: To be honest I forgot to even check.

JimBobSquarePants:
Sorry! Completely missed that was internal!

src/SixLabors.Fonts/FontRectangle.cs
```diff
@@ -0,0 +1,374 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FontRectangle"/> struct.
+        /// </summary>
+        /// <param name="point">
+        /// The <see cref="Point"/> which specifies the rectangles point in a two-dimensional plane.
+        /// </param>
+        /// <param name="size">
+        /// The <see cref="Size"/> which specifies the rectangles height and width.
+        /// </param>
+        internal FontRectangle(Vector2 point, Vector2 size)
```
dlemstra: It looks like we don't need this as an internal constructor? Maybe make it private then instead?

antonfirsov:
So we completely, and irreversibely made up our mind regarding this new structure? (Considering the drawbacks I pointed out).

Personally I'm still like:
image

src/SixLabors.Fonts/FontRectangle.cs
tocsoft: your right, made it private.

JimBobSquarePants:
@tocsoft usually spots issues long before I do so if he is confident then I am.

JimBobSquarePants:
@antonfirsov I say leave the issues for now. The library needs a proper going over to fix up warnings. Any missing tests can be added then.

We should however raise issues to ensure the work gets done.

PR #122, opened 2020-01-18
install TTC files

### LINKED ISSUE #120: Add support for reading font collection files (with the .ttc extension).
Currently fonts like the simsun.ttc font cannot be loaded. Would it be much work to implement this?

Code:

```
SixLabors.Fonts.FontDescription.LoadDescription(@"C:\Windows\Fonts\simsun.ttc");
```

tocsoft:
Adds installing ttc files with tests and support for SystemFontCollection.

This merges @woutware's PR #121 and tweeks the API surface.

Fixes #120 
Resolves #121


woutware:
Hi @tocsoft , thanks for cleaning up the code and adding the tests, wasn't sure how to setup the tests to fit in with the rest. I just have 2 remarks/questions:

If I understand correctly, the lazy font loading will just start reading a font at specified offset, independently from other fonts in the collection, and therefore in memory the collection glyf table is not actually reused, but will be loaded multiple times? I don't particularly care about this potentially inefficiency, I'm just wondering if I'm understanding it correctly. Maybe it's worth adding a comment about in the code.

Minor curiosity: why return `HashSet<FontFamily>()`? For just a few elements an array is more efficient, plus it's being returned as an ienumerable anyway.

tocsoft:
your correct we are not actually trying to share tables in memory at this time, but its a candidate for future optimizations if its deemed required.
 
HashSet is being used because i'm only after only the distinct set of font families for each instance without me having to think too hard on the most efficient way of creating a distinct set.

woutware:
Ah makes sense, hadn't thought about that they could not be distinct.

woutware:
When using the FontReader for all the fonts it will reuse the tables, so perhaps when triggering loading one font out of the collection, it could force loading all the fonts in that collection.

tocsoft:
`FontReader` only knows how to read a single font as it reads the font header during construction at which point everything is locked into that it would require a larger refactor to have a collection reader that can store tables and allow lookup no only by tag but also by offset... it could be done just none trivially

PR #123, opened 2020-01-18
migrate to github actions and fix warnings

tocsoft:


tocsoft:
merging this just moving over to github actions like imagesharp

JimBobSquarePants:
Heads up. GitVersion.yml lookup is case sensitive. It’s ignoring our config

PR #126, opened 2020-02-24
Provide CultureInfo down stack to allow loading metadata in specific culture

tocsoft:
### Prerequisites

### Description
This add the ability to provide a `CultureInfo` when querying a `FontCollection` to enable searching font metadata (font name, family, description etc) in a cultural aware manor. This allows fonts with localised value to be loadable by their localised name.

Culture aware APIs are only exposed on netstandard 2.1 compatible platforms (we don't have the internal APIs available to support cultures on earlier platforms).

APIs that have a culture aware counterpart all act as though they the same as calling the culture aware ones with the Invariant Culture specified.

Reason we use default to Invariant Culture is for variant compatibility.. otherwise behaviour may change when moving from netcoreapp 2.0 to netcoreapp 3.1.


src/SixLabors.Fonts/FontDescription.cs
```diff
@@ -27,8 +28,9 @@ internal FontDescription(string fontName, string fontFamily, string fontSubFamil
         /// <param name="nameTable">The name table.</param>
         /// <param name="os2">The os2.</param>
         /// <param name="head">The head.</param>
-        internal FontDescription(NameTable nameTable, OS2Table? os2, HeadTable? head)
-            : this(nameTable.FontName, nameTable.FontFamilyName, nameTable.FontSubFamilyName, ConvertStyle(os2, head))
+        /// <param name="culture">The culture to load metadata in.</param>
+        internal FontDescription(NameTable nameTable, OS2Table? os2, HeadTable? head, CultureInfo culture)
+            : this(nameTable.FontName(culture), nameTable.FontFamilyName(culture), nameTable.FontSubFamilyName(culture), ConvertStyle(os2, head))
```
BSG-75: I think we could also have a `NameTable` property with `internal` access level, so `FontCollection` can access various valus from `NameTable`, through `FontDescription`.

src/SixLabors.Fonts/FontCollection.cs
```diff
@@ -15,13 +16,24 @@ namespace SixLabors.Fonts
     /// </summary>
     public sealed class FontCollection : IFontCollection
     {
-        private readonly Dictionary<string, List<IFontInstance>> instances = new Dictionary<string, List<IFontInstance>>(StringComparer.OrdinalIgnoreCase);
-        private readonly Dictionary<string, FontFamily> families = new Dictionary<string, FontFamily>(StringComparer.OrdinalIgnoreCase);
+        private readonly Dictionary<string, List<IFontInstance>> instances = new Dictionary<string, List<IFontInstance>>(StringComparer.CurrentCultureIgnoreCase);
+        private readonly Dictionary<string, FontFamily> families = new Dictionary<string, FontFamily>(StringComparer.CurrentCultureIgnoreCase);
+        private readonly CultureInfo culture;
```
BSG-75: And if LCID is supported (for example in .net standard 2.0+), then we could also add a `Dictionary<int, Dictionary<string, FontFamily>> familiesByLanguages`  where `int` is an `LCID`, to add every font family name in every language. These entries are already inside `NameTable` so I think probably it's better to make font families names available to the public instead of keeping them private whenever possible. Each value of `familiesByLanguages` (i.e. each `Dictionary<string, FontFamily>`) uses its own `StringComparer` in its own culture.

When a search begins, firstly we search in the preferred culture (i.e. `FontCollection.families`). If we didn't find anything,  (and if LCID is supported, i.e. we are .net standard 2.0+ and we have the above `FontCollection.familiesByLanguages`), then we iterate through `FontCollection.familiesByLanguages` and search for the font family in every possible language.

src/SixLabors.Fonts/IReadonlyFontCollection.cs
```diff
@@ -12,13 +14,39 @@ namespace SixLabors.Fonts
     public interface IReadOnlyFontCollection
     {
         /// <summary>
-        /// Gets the collection of <see cref="FontFamily"/> objects associated with this <see cref="FontCollection"/>.
+        /// Gets the collection of <see cref="FontFamily"/> objects associated with this <see cref="FontCollection"/> in the current threads culture.
```
JimBobSquarePants, line 17: invariant culture no?

src/SixLabors.Fonts/FontCollection.cs
tocsoft: decided to go with make users specify culture when searching they want to use culture specific names

PR #130, opened 2020-03-17
Updated documentation for SystemFonts.Find

Turnerj:
### Prerequisites

### Description
Just a small update to the documentation to match the actual internal logic.

### DIFF
--- a/src/SixLabors.Fonts/SystemFonts.cs
+++ b/src/SixLabors.Fonts/SystemFonts.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Six Labors and contributors.
+// Copyright (c) Six Labors and contributors.
 // Licensed under the Apache License, Version 2.0.
 
 using System;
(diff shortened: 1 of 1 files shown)


src/SixLabors.Fonts/SystemFonts.cs
```diff
@@ -33,15 +33,16 @@ public static class SystemFonts
         /// Finds the specified font family from the system font store.
         /// </summary>
         /// <param name="fontFamily">The font family.</param>
-        /// <returns>The family if installed otherwise null</returns>
+        /// <returns>The found family.</returns>
```
JimBobSquarePants: Thanks for doing this. Really appreciate it! Just one tiny typo.  found => font.

Turnerj: No problem at all! I actually chose "The found family" over "The font family" as the `TryFind` method below refers to the out-variable as "The found family". Happy to still change it (or both if you like) 🙂

JimBobSquarePants: Oh... ok, yeah let's leave it as-is for now then.

PR #132, opened 2020-04-05
Allow falling back to alternative font families if glyph is missing

### LINKED ISSUE #117: fallback fonts for when glyph is not found
I have a use case where I need to draw text in an image. This text is user input and can be a whole range of languages, It can even be a mix of 2 languages. For example english and chinese.

It would be nice that we can list fallback font's so that if a glyph for a unicode character is not found then it tries in the next provided font until it finds the glyph or there are no more fonts.

tocsoft:
### Prerequisites

### Description
Provides addition APIs on the `RenderOptions` class that allows for specific alternative font families for falling back to if the currently provided font doesn't contain the have the required glyph.

Resolves #117

### DIFF
--- a/src/SixLabors.Fonts/AppliedFontStyle.cs
+++ b/src/SixLabors.Fonts/AppliedFontStyle.cs
@@ -1,15 +1,47 @@
-﻿// Copyright (c) Six Labors and contributors.
+// Copyright (c) Six Labors and contributors.
 // Licensed under the Apache License, Version 2.0.
+
+using System.Collections.Generic;
+using System.Numerics;
 
 namespace SixLabors.Fonts
 {
     internal struct AppliedFontStyle
     {
-        public IFontInstance Font;
+        public IEnumerable<IFontInstance> FallbackFonts;
+        public IFontInstance MainFont;
         public float PointSize;
         public float TabWidth;
         public int Start;
         public int End;
         public bool ApplyKerning;
+
+        public GlyphInstance GetGlyph(int codePoint)
+        {
+            GlyphInstance glyph = this.MainFont.GetGlyph(codePoint);
+            if (glyph.Fallback)
+            {
+                foreach (var f in this.FallbackFonts)
+                {
+                    var g = f.GetGlyph(codePoint);
+                    if (!g.Fallback)
+                    {
+                        return g;
+                    }
+                }
+            }
+
+            return glyph;
+        }
+
+        public Vector2 GetOffset(GlyphInstance glyph, GlyphInstance previousGlyph)
+        {
+            if (glyph.Font != previousGlyph?.Font)
+            {
+                return Vector2.Zero;
+            }
+
+            return ((IFontInstance)glyph.Font).GetOffset(glyph, previousGlyph);
+        }
     }
 }
--- a/src/SixLabors.Fonts/AppliedFontStyle.cs
+++ b/src/SixLabors.Fonts/AppliedFontStyle.cs
@@ -8,7 +8,7 @@ namespace SixLabors.Fonts
 {
     internal struct AppliedFontStyle
     {
-        public IEnumerable<IFontInstance> FallbackFonts;
+        public IFontInstance[] FallbackFonts;
         public IFontInstance MainFont;
         public float PointSize;
         public float TabWidth;
--- a/src/SixLabors.Fonts/RendererOptions.cs
+++ b/src/SixLabors.Fonts/RendererOptions.cs
@@ -1,6 +1,8 @@
 // Copyright (c) Six Labors and contributors.
 // Licensed under the Apache License, Version 2.0.
 
+using System.Collections.Generic;
+using System.Linq;
 using System.Numerics;
 
 namespace SixLabors.Fonts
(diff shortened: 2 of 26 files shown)


Tylertron1998:
Would love to see this merged - I've wanted to switch to ImageSharp for a while, but this has been holding me back!

src/SixLabors.Fonts/AppliedFontStyle.cs
```diff
@@ -1,15 +1,47 @@
...
         public int Start;
         public int End;
         public bool ApplyKerning;
+
+        public GlyphInstance GetGlyph(int codePoint)
+        {
+            GlyphInstance glyph = this.MainFont.GetGlyph(codePoint);
+            if (glyph.Fallback)
+            {
+                foreach (var f in this.FallbackFonts)
```
JimBobSquarePants, line 24: How do we ensure here that re-enumerating the enumerable here is not costly?

src/SixLabors.Fonts/RendererOptions.cs
```diff
@@ -10,14 +12,15 @@ namespace SixLabors.Fonts
     /// </summary>
     public sealed class RendererOptions
     {
+        private static FontFamily[] emptyFamilyCollection = new FontFamily[0];
```
JimBobSquarePants, line 15: `Array.Empty<FontFamily>()`

src/SixLabors.Fonts/AppliedFontStyle.cs
tocsoft: its only ever called once.

src/SixLabors.Fonts/RendererOptions.cs
```diff
@@ -71,11 +73,81 @@ public RendererOptions(Font font, float dpi, Vector2 origin)
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RendererOptions"/> class.
+        /// </summary>
+        /// <param name="font">The font.</param>
+        /// <param name="dpiX">The X dpi.</param>
+        /// <param name="dpiY">The Y dpi.</param>
+        /// <param name="origin">The origin location.</param>
+        /// <param name="fallbackFontFamilies">An optional collection of font families to fallback to if the glyph is missing from the <paramref name="font"/>.</param>
+        public RendererOptions(Font font, float dpiX, float dpiY, Vector2 origin, IEnumerable<FontFamily> fallbackFontFamilies)
```
JimBobSquarePants, line 144: I think you're much better removing all these constructor overloads and using default property assignment.

src/SixLabors.Fonts/RendererOptions.cs
tocsoft: yep much better totally forgot that one existed

src/SixLabors.Fonts/AppliedFontStyle.cs
tocsoft: skip that wrong file

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -85,18 +85,27 @@ public IReadOnlyList<GlyphLayout> GenerateLayout(ReadOnlySpan<char> text, Render
...
                 }
 
-                if (spanStyle.Font.LineHeight > unscaledLineHeight)
+                bool hasFourBytes = char.IsHighSurrogate(text[i]);
+                int codePoint = hasFourBytes ? char.ConvertToUtf32(text[i], text[i + 1]) : text[i];
+
+                GlyphInstance glyph = spanStyle.GetGlyph(codePoint);
+                if (glyph == null)
+                {
+                    throw new Exceptions.FontException($"Cannot find a glyph for code point {codePoint}");
```
JimBobSquarePants, line 94: We should use a throwhelper in a loop. This might be inlinable.

src/SixLabors.Fonts/RendererOptions.cs
tocsoft: hmm... nether `Vector2` or `IEnumerable<FontFamily>` can be default parameters especially as we won't want to enable c#8 nullable referenece types and don't want to expose the arguments as nullable... I will however remove the fallback arguments and move them over to a setable property.

JimBobSquarePants: Couldn't you use `Vector2?` and `IReadOnlyCollection<T>` for the properties?

We use `PointF?` in `ResizOptions`.

src/SixLabors.Fonts/AppliedFontStyle.cs
tocsoft: sorted... now using an Array not IEnumerable

src/SixLabors.Fonts/RendererOptions.cs
tocsoft: 1. we could do but we shouldn't be encouraging magic null values in my opinion and 2. I would rather not break the API at this stage without good reason, these constructors are already in use by imagesharp and other third parties.

src/SixLabors.Fonts/TextLayout.cs
tocsoft: sorted

PR #134, opened 2020-04-25
Refactor glyph layer rendering

tocsoft:
- fix possible text measuring of color glyphs substitutes
- ensure correct text advance used when multiple layers with potentially different advances
- custom emoji font build which fixes the advance widths in hmtx table

### Prerequisites

### Description

While trying to implement emoji support into Drawing I hit a couple of issues that required a small amount of refactor inside the layout/renderer.

Enables support for measuring fonts with emoji glyphs too.


JimBobSquarePants:
What are the `**Copy` classes for?

tocsoft:
They are all copies of classes in `ImageSharp.Drawing` with the tweaks needed to make it render emoji correctly, basically I started in drawing making the required changes that are in the `**copy` classes found we had issues so I back ported them over to fonts to carry on and verify the end to end will actually work before raising the PR. 

Turns out I had to copy over more stuff that I wanted as you can't actually call the `apply()` method on a brush applicator outside the drawing project thus making it so i had to copy even more stuff that I wanted to. 

But in conclusion they aren't in the SixLabors.Fonts package and just in a a bit of a scratch console app to help debug the rest of the library... so feel free to ignore them.

src/SixLabors.Fonts/RendererOptions.cs
```diff
@@ -141,6 +141,11 @@ public RendererOptions(Font font, float dpiX, float dpiY, Vector2 origin)
         /// </summary>
         public Vector2 Origin { get; set; } = Vector2.Zero;
 
+        /// <summary>
+        /// Gets or sets weather we enable various color font formats.
```
JimBobSquarePants, line 145:
```suggestion
        /// Gets or sets a value indicating whether we enable various color font formats.
```

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -161,7 +162,18 @@ public IReadOnlyList<GlyphLayout> GenerateLayout(ReadOnlySpan<char> text, Render
                         location.X = glyphLocation.X;
                     }
 
-                    layout.Add(new GlyphLayout(codePoint, new Glyph(glyph, spanStyle.PointSize), glyphLocation, glyphWidth, glyphHeight, lineHeight, startOfLine, false, false));
+                    foreach (var g in glyphs)
+                    {
+                        var w = (g.AdvanceWidth * spanStyle.PointSize) / scale;
```
JimBobSquarePants, line 167: Is the Fonts library using the same standards as others? I can't tell what type w without loading up the IDE.

tocsoft: should be... its using the 'shared-infrastructure' repo.

It does however have a subtly different .editorconfig so that could be the cause of the differences.

JimBobSquarePants: Yeah that makes sense. There's some known bugginess in the way editorconfig determines when `var` should be used with built in types. The config from the shared infrastructure contains the workaround. I've set ImageSharp to automatically copy taht file and the gitattributes file to the root.

PR #146, opened 2020-07-17
migrate to minver from gitversion

tocsoft:
complete the set migrating from gitversion to minver


JimBobSquarePants:
Looks good. You just need to update the reference imagesharp version in DrawWithImageSharp.

PR #153, opened 2020-12-12
Add LineSpacing property to allow manipulating the line height

### LINKED ISSUE #152: Parameter to control text line height/leading
Follow-up on SixLabors/ImageSharp.Drawing#112:

We need to introduce a parameter in title in [RenderOptions](https://github.com/SixLabors/Fonts/blob/4beaff16420c706a10c293cca6e1c1037a420d36/src/SixLabors.Fonts/RendererOptions.cs).

That parameter should then somehow alter the following `offset` value:
https://github.com/SixLabors/Fonts/blob/4beaff16420c706a10c293cca6e1c1037a420d36/src/SixLabors.Fonts/TextLayout.cs#L279

/cc @tocsoft @JimBobSquarePants

JimBobSquarePants:
### Prerequisites

### Description
Add a new property `RenderOptions.LineSpacing` with a default value of `1`. This is applied as a multiplier of the lineheight in the same manner as MS Word. Fixes #152 

Left is default, Right is with a value of `2`.

Comparison of line spacing

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -32,11 +32,6 @@ jobs:
                   git fetch --prune --unshallow
                   git submodule -q update --init --recursive
 
-            - name: Setup DotNet SDK
-              uses: actions/setup-dotnet@v1
-              with:
-                  dotnet-version: "3.1.x"
-
             - name: Pack
               shell: pwsh
               run: ./ci-pack.ps1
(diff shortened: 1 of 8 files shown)


antonfirsov (dismissed):
Looks good, just make sure existing drawing tests pass with the new code!

antonfirsov:
We need regression tests for the feature, either here or in the drawing repo (or both).

I'm not sure about practices in Fonts.Tests, but - if possible - I would try to validate distances between polygon centers after rendering `.\n\.\n.` for example.

tocsoft (dismissed):
lgtm... the build issues will be because we changed the return types from Drawing.

These APIs will only be being used in sample apps not the main code.

JimBobSquarePants:
Yep. Forgot to push the sample file I changed. Will push tonight

PR #156, opened 2021-01-02
Implement Unicode Line Breaking Algorithm

JimBobSquarePants:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/Fonts/pulls) open
- [x] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description

Provides an implementation of the Unicode Line Breaking Algorithm UAX # 14
https://www.unicode.org/reports/tr14/tr14-37.html

- [x] Port Unicode Trie, Builder + Tests
- [x] Port Line Breaker implementation + Tests
- [x] Wire up to `TextLayout` algorithm
- [x]  Cleanup and optimization.

The port is based upon several sources indicated in the readme and passes over 7500 tests. 

There are ~70 outstanding tests which I have individually checked and they appear to be expecting the incorrect result according to the specification. This is inline with other implementations.

@tocsoft I'll have to leave you to do the actual wiring up of the `LineBreaker` as I will probably make a mistake. Here's how [RickTextKit](https://github.com/toptensoftware/RichTextKit/blob/a2733e43ba8484032adfb1defbdde41c09986e7f/Topten.RichTextKit/TextBlock.cs#L1312) does it.

Sorry about the additional noise but I was getting collisions with the CLR `BinaryReader` class.

Fixes #103

Issue #66, opened 2018-05-22
Support for RTL text


tocsoft:
Add option to text layout engine to support rendering 'right to left' instead of the current 'left to right'.

This will require new option on `RendererOptions` to allow configuring 'right to left` 
and updating https://github.com/SixLabors/Fonts/blob/4ee628bad5c0206440bd6cc865ee3d6a3285c329/src/SixLabors.Fonts/TextLayout.cs#L25 to layout the characters in the corrent 'right to left' style.

JimBobSquarePants:
While I've added a Bidi implementation in #160 I'm really not happy with the shape of the API it brings. 

I want to investigate this alternative.
https://github.com/Tehreer/SheenBidi

[LayoutFarm Typography](https://github.com/LayoutFarm/Typography) is currently using a C# version that appears to be based on https://archive.codeplex.com/?p=sheenbidi written by the same author as the C version.

PR #160, opened 2021-01-17
WIP Add UAX #9: Unicode Bidirectional Algorithm Support

JimBobSquarePants:
### Prerequisites

### Description
- [x] Port BidiTrie generation 
- [x] Port Bidi type and dependencies
- [x] Port BidiData type
- [x] Port ICU Tests
- [ ] Wire up Bidi implementation to `TextLayout`
- [ ] Additional Tests and performance tweaks.

This adds an implementation of the [UAX #9: Unicode Bidirectional Algorithm Support](https://unicode.org/reports/tr9) supporting Unicode 13 which should go a long way towards helping us support RTL languages #66 . 

Most of the work has been ported from [RichTextKit](https://github.com/toptensoftware/RichTextKit). I've kept the basic shape of their implementation but used my own versions of their `Buffer<T>`, `Slice<T>`, etc types for better performance. 

All of the ICU tests pass (_770241 - Bidi, 91699 Bidi Character_) so we have a good start.

As before with #156 I'll need @tocsoft to wire it all up. [`TextBlock`](https://github.com/toptensoftware/RichTextKit/blob/a2733e43ba8484032adfb1defbdde41c09986e7f/Topten.RichTextKit/TextBlock.cs) should help there. 

Feel free to update whatever you need to API wise to fit requirements. I've just done a (_fairly_) dumb port.


tocsoft:
@JimBobSquarePants getting this integrated into the rendering/layout pipeline won't be a simple/quick as the line break one.... I had a look over the `TestBlock` stuff in RichTextKit and its making me think that instead of integrating this stuff into our current layout pipeline instead I wonder if we could instead take the whole of their layout engine and just swap out their paint method to target our `IGlyphRenderer` as our target instead of `SKCanvas`.

JimBobSquarePants:
>I wonder if we could instead take the whole of their layout engine and just swap out their paint method to target our `IGlyphRenderer` as our target instead of `SKCanvas`.

@tocsoft I am more than happy to take your lead on this! I can port a few more things if you like to make it easier? `GraphemeCluster`, `WordBoundary`?

JimBobSquarePants:
@tocsoft what I’ll do with this then is update coverage tests and mark it ready for review. I’ll then migrate the code for the parts mentioned above in a single PR so we can do a final PR to wire it all up to an implementation based on the source from that repo.

JimBobSquarePants:
Ok @tocsoft I think everything required has been ported across. Over to you.

JimBobSquarePants:
Actually... Give me a day. I want to do some refactoring work to give you a `CodePoint` struct and an easy way to read it from a string which will contain all the relevant info we need.

JimBobSquarePants:
OK... I've added I whole bunch of features including a new `CodePoint` type you can use to query Unicode data against. The API is shaped very much like the new `Rune` type in the BCL.

Most of the parsing code can read directly from the string without allocating but I kept breaking (_pun intended_) the `LineBreakerAlgorithm` when I tried to convert it to do the same. I can have another crack if necessary though.

For the most part, everything works the same way as RichTextKit but allocates less and should be faster.

I think I've reached my limit now until we start to get this wired up properly. I can reshape the APIs to match requirements if need be.

Over to  you @tocsoft 😄

JimBobSquarePants:
Will need to convert Grapheme and Bidi algorithms to use `ReadOnlySpan<char>`. Should be doable.

jeffska:
I've dealt with this stuff a lot, so I'm going to be "that guy" for a bit. Other than wanting to stay fully managed, is there a reason to just not take the dependency on ICU and be done with it? In my experience you're still going to need a Unicode line break algorithm and glyph shaping. For the latter, that's probably going to mean a dependency on HarfBuzz/HarfBuzzSharp; unless you're also planning to write a fully managed glyph shaping library. 

But, if that's the plan, and it can get to ICU/HarfBuzz performance levels, then I would love to ditch those dependencies.

I'll also see if I can give this a look and maybe try to plug it in to my code and do some comparisons w/ the ICU implementation.

JimBobSquarePants:
Thanks for having a look @jeffska 

Yeah, basically we eventually want to have everything running together in this library to pass on to ImageSharp.Drawing. It's a big job but it's imperative we have everything in managed code where we can optimize the guts to suit our needs and have super easy deployment. 

I've just bumped the linebreak algorithm impl code so that all the tests pass. It can run all 7652 tests in ~90ms which includes loading/parsing the file to build the tests so it's not shabby. 

The Bidi code is an area I'm not too happy with. I copied that from another solution and I think we can tighten it up.

JimBobSquarePants:
Things I'd like to do:

- Convert `BidiData` to use `ReadOnlySpan<char>`. Bit of a chicken and egg scenario here. I can't do that until we start wiring it up as I don't fully understand how it works. Would be great to turn it into an enumerator.  
- Simplify `LineBreakAlgorithm` so it's an enumerator. (I can do that now but currently keeping the API the same as RTK)
- Simplify `GraphemeAlgorithm` so it's an enumerator and return only the graphemes not the breaks to match MS and Avalon implementations. (I can do that now but currently keeping the API the same as RTK).

I've had several attempts at integration now but I simply don't know enough about the space to create the text layout.

jeffska:
I grabbed the branch and tried integrating it w/ my text layout routines in place of my ICU dependencies. It did what I needed, so that's good. I obviously had to make some things public to get it to work. I was using ICU for line breaking (have an iterator that wraps ICU calls), and BiDi determination. Because of the way I create my text runs, I don't need full BiDi levels, just determination for the single base direction. I was using the ICU ubidi_getBaseDirection which returns LTR or RTL based on the first strong type it finds in the text. I emulated that iterating through the text and calling CodePoint.GetBidiType.

The only thing that this doesn't provide is reordering to display order for BiDi. I had earlier code that used ICU ubidi_getVisualMap for that, but switched to letting HarfBuzzSharp handle it, which is what I believe RTK does.

So if this gets finished, and exposes the right stuff publicly, I can at least get rid of my ICU dependency.

JimBobSquarePants:
@jeffska What does `ubidi_getVisualMap` do? I feel like this (_reordering to display order for BiDi_) is part of the wiring up that still needs to happen which I will need help with.

JimBobSquarePants:
@tocsoft I think we should review and merge what we have currently and handle wiring up the Bidi stuff at a later date otherwise this will grow out of hand.

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -148,32 +146,31 @@ public IReadOnlyList<GlyphLayout> GenerateLayout(ReadOnlySpan<char> text, Render
...
+                    nextWrappableLocation = b.PositionWrap - 1;
+                    nextWrappableRequired = b.Required;
                 }
 
                 float glyphWidth = glyph.AdvanceWidth * spanStyle.PointSize / scale;
                 float glyphHeight = glyph.Height * spanStyle.PointSize / scale;
 
-                if (codepoints[i] != '\r' && codepoints[i] != '\n' && codepoints[i] != '\t' && codepoints[i] != ' ')
+                // TODO: !CodePoint.IsWhitespace(codePoint) && ! CodePoint.IsBreak(codePoint) ??
+                if (codePoint.Value != '\r' && codePoint.Value != '\n' && codePoint.Value != '\t' && codePoint.Value != ' ')
```
JimBobSquarePants, line 173: @tocsoft I'm curious about the explicit char rules here. This seems like something we can determine with our new Unicode tooling.

JimBobSquarePants: Fixing via https://github.com/SixLabors/ImageSharp.Drawing/issues/131

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -207,54 +203,80 @@ public IReadOnlyList<GlyphLayout> GenerateLayout(ReadOnlySpan<char> text, Render
...
+                        location.Y += lineHeight;
+                        totalHeight += lineHeight;
+                        firstLine = false;
+                        lastWrappableLocation = -1;
                     }
 
                     previousGlyph = glyph;
                 }
-                else if (codepoints[i] == '\r')
+                else if (codePoint.Value == '\r')
```
JimBobSquarePants, line 248: I need to review the logic here. \r on it's own could potentially be incorrectly detected on it's own since we enumerating the code points.

JimBobSquarePants:
I'm gonna merge this. There's a few fixes I'd like to get out into a new beta. We can deal with actual Bidi integration separately.

PR #184, opened 2021-07-16
Use ImageSharp.Drawing package for samples.

JimBobSquarePants:
### Prerequisites

### Description

Updated the samples project to use the text drawing processor from ImageSharp.Drawing in combination with the Fonts project. This lets us more easily identify potential issues (_I actually found and fixed two doing this_).


tocsoft (approved):
Generally happy with this, we have actually flip flopped between using the Rendered from ImageSharp.Drawing and using a custom processor. I believe the current reason it switched to a custom implementation was when I added emojji/color font support I needed to validate the API changes/additions where sufficient before we could make the change to IS.Drawing proper.

JimBobSquarePants:
Yeah, the way I worked with it to validate and fix issues was to replace the NuGet ref with my local drawing project. It’s a bit convoluted but enlightening wise for issue discovery.

PR #188, opened 2021-07-27
Add support for reading woff2 fonts

brianpopow:
### Prerequisites

### Description

This **PR** adds support for reading fonts in the woff2 format. Spec can be found here: [Woff2](https://www.w3.org/TR/WOFF2) 

This is based on the implementation from [Typography](https://github.com/LayoutFarm/Typography)

~~Proper unit tests are still missing.~~


JimBobSquarePants:
Awesome!

src/SixLabors.Fonts/Tables/General/ColrTable.cs
```diff
@@ -21,7 +21,7 @@ public ColrTable(BaseGlyphRecord[] glyphRecords, LayerRecord[] layers)
 
         public static ColrTable? Load(FontReader reader)
         {
-            using (var binaryReader = reader.TryGetReaderAtTablePosition(TableName))
+            using (BigEndianBinaryReader? binaryReader = reader.TryGetReaderAtTablePosition(TableName))
```
JimBobSquarePants, line 24: Ooh... Hadn't spotted that before. Should be this IMO.

```c#
bool FontReader.TryGetReaderAtTablePosition(TableName, out BigEndianBinaryReader? reader)
```

src/SixLabors.Fonts/Tables/General/Glyphs/GlyphVector.cs
```diff
@@ -23,19 +36,91 @@ internal GlyphVector(Vector2[] controlPoints, bool[] onCurves, ushort[] endPoint
...
+            float newYmax = 0;
+
+            Vector2[] glyphPoints = this.ControlPoints;
+            for (int i = 0; i < glyphPoints.Length; ++i)
+            {
+                Vector2 p = glyphPoints[i];
+                float x = p.X;
+                float y = p.Y;
+
+                glyphPoints[i] = new Vector2((float)Math.Round((x * m00) + (y * m10)), (float)Math.Round((x * m01) + (y * m11)));
```
JimBobSquarePants, line 96: I'd double check we can't just use the following here. 

https://docs.microsoft.com/en-us/dotnet/api/system.numerics.vector2.transformnormal?view=netframework-4.8

brianpopow:
what would be the right way add a unit test to verify all glyphs are loaded correctly? Do I need to render the glyphs for that? Is there a example unit test i could use for reference?

JimBobSquarePants:
Not sure on the tests tbh, it’s only recently I’ve been focusing on the non-Unicode stuff.

I see you’ve added some tests similar to the glyph metrics ones I recently did. Ideally we’d be able to render the glyphs. I think, because versioning will be in our favour we should be able to reference the drawing library in the tests but use the project reference and actually draw som reference output.

tocsoft:
I wouldn't rasterise the font for this sort of testing.

You'll be able to use the https://github.com/SixLabors/Fonts/blob/8e4d18550d96c2ba5932a6b3c897db72eb44691c/tests/SixLabors.Fonts.Tests/ColorGlyphRenderer.cs#L8 to record the rendering process (rather than rasterizing) then just validate the recorded points. Just use `TextRenderer` and pass in the test Glyph/ColorGlyphRenderer and you should be good to go.

You should be able to just validate the engine against itself ... i.e. record rendering the same font via loading a ttf and a woff2 version side by side and validate there outputs are identical its self consistency we are after here after all.

JimBobSquarePants:
Excellent. Thanks @tocsoft !!

src/SixLabors.Fonts/Tables/General/Glyphs/GlyphVector.cs
brianpopow: please review, if that looks correct: fba5f84d

src/SixLabors.Fonts/Tables/General/ColrTable.cs
brianpopow: changed with fbcfa78d

src/SixLabors.Fonts/Tables/General/Glyphs/GlyphVector.cs
JimBobSquarePants: Will do!

src/SixLabors.Fonts/Tables/General/Glyphs/GlyphVector.cs
```diff
@@ -1,41 +1,123 @@
...
+        {
+            Vector2[] glyphPoints = this.ControlPoints;
+            for (int i = glyphPoints.Length - 1; i >= 0; --i)
+            {
+                glyphPoints[i] = Offset(glyphPoints[i], dx, dy);
+            }
+
+            Bounds orgBounds = this.Bounds;
+            this.Bounds = new Bounds(
+                (short)(orgBounds.Min.X + dx),
```
JimBobSquarePants, line 62: Why the cast to `short` here?

src/SixLabors.Fonts/Tables/Woff2Utils.cs
```diff
@@ -0,0 +1,767 @@
...
+
+        private static byte[] ExpandBitmap(byte[] orgBBoxBitmap)
+        {
+            byte[] expandArr = new byte[orgBBoxBitmap.Length * 8];
+
+            int index = 0;
+            for (int i = 0; i < orgBBoxBitmap.Length; ++i)
+            {
+                byte b = orgBBoxBitmap[i];
+                expandArr[index++] = (byte)((b >> 7) & 0x1);
```
JimBobSquarePants, line 656: This looks like a potentially vectorizable method.

JimBobSquarePants (approved):
I'll leave the final review to @tocsoft but I couldn't find anything of concern in the PR. 

Super impressed by the speed you've done this. I've dipped into the Typography code a few time and struggled to navigate it.

src/SixLabors.Fonts/Tables/General/Glyphs/GlyphVector.cs
brianpopow: not needed, removed with 6483f949

tocsoft (commented):
Generally fine with this.. well done, only small thing is the introduction/change in pattern for dealing with table names `const`s. It feels like moving them all together doesn't give you much except a lot of code churn and the ability to have table names without the corresponding parsing/loading logic. I would personally remove the `TableNames` class and just make the preexisting `TableName` Consts on each table class public/internal... but if there is a good reason to have them all together that I'm missing then we should remove the `TableName` const for the table classes and just use the combined set... we shouldn't have this hybrid/mixed model.

src/SixLabors.Fonts/Tables/General/CMapTable.cs
```diff
@@ -12,14 +12,14 @@ namespace SixLabors.Fonts.Tables.General
     [TableName(TableName)]
     internal sealed class CMapTable : Table
     {
-        private const string TableName = "cmap";
+        private const string TableName = TableNames.Cmap;
```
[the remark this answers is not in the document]
brianpopow: For some reason, I thought those table names are common to more formats and are not specific to woff2, but its only a few of them. Your right the change was not good, I have reverted it and made the table names internal now.

JimBobSquarePants (approved):

PR #190, opened 2021-08-09
Expose methods to get paths for fonts.

JimBobSquarePants:
### Prerequisites

### Description

Adds two new methods to the API to allow attempts to retrieve a filesystem path for a given `Font` or collection for a `FontFamily`.

`FontFamily`

```c#
/// <summary>
/// Gets the collection of filesystem paths to the font family sources.
/// </summary>
/// <param name="paths">
/// When this method returns, contains the filesystem paths to the font family sources,
/// if the path exists; otherwise, an empty value for the type of the paths parameter.
/// This parameter is passed uninitialized.
/// </param>
/// <returns>
/// <see langword="true" /> if the <see cref="FontFamily" /> was created via filesystem paths; otherwise, <see langword="false" />.
/// </returns>
public bool TryGetPaths(out IEnumerable<string> paths)
```

`Font`

```c#
/// <summary>
/// Gets the filesystem path to the font family source.
/// </summary>
/// <param name="path">
/// When this method returns, contains the filesystem path to the font family source,
/// if the path exists; otherwise, the default value for the type of the path parameter.
/// This parameter is passed uninitialized.
/// </param>
/// <returns>
/// <see langword="true" /> if the <see cref="Font" /> was created via a filesystem path; otherwise, <see langword="false" />.
/// </returns>
public bool TryGetPath([NotNullWhen(true)] out string? path)
```

In addition I performed a cleanup and refactor of the `IFontCollection` and `IReadonlyFontCollection` APIs. Functionally they remain the same.

Notably:
- Converted `Fontfamily` into a struct since it was using value type equality operators.
- Renamed `Find**` method signatures to `Get**` since we throw when no match is available and we do not expose a predicate.
- Renamed `Install**` method signatures to `Add**` since we are not installing anything in the OS.
- Replaced a single instance of `CultureInfo.CurrentCulture` with `CultureInfo.InvariantCulture` in `IFontCollection.Add(..)` to make it consistant with the other `Add` methods.
- Introduced `IReadOnlyFontMetricsCollection` which is implemented (_but hidden_) in `FontCollection` and passed via the internal constructor to `FontFamily`
- Flattened the namespace for `Exception` types.


samples/ListFonts/Program.cs
```diff
@@ -7,25 +7,30 @@
...
-                Console.Write(f.Name.PadRight(len));
-                Console.Write('\t');
-                Console.Write(string.Join(",", f.AvailableStyles.OrderBy(x => x).Select(x => x.ToString())));
-                Console.WriteLine();
+                IOrderedEnumerable<FontStyle> styles = family.GetAvailableStyles().OrderBy(x => x);
+                foreach (FontStyle style in styles)
+                {
+                    Font font = family.CreateFont(0F, style);
+                    font.TryGetPath(out string path);
+                    pairings.Add(new Pairing(font.Name, path));
```
tocsoft, line 26: a simple named tuple was probably all you needed here instead of creating the full struct, but not particually important and isn't really an issue I would flag up to change.

src/SixLabors.Fonts/Font.cs
```diff
@@ -106,38 +114,81 @@ public Font(Font prototype, float size)
...
+        /// This parameter is passed uninitialized.
+        /// </param>
+        /// <returns>
+        /// <see langword="true" /> if the <see cref="Font" /> was created via a filesystem path; otherwise, <see langword="false" />.
+        /// </returns>
+        public bool TryGetPath([NotNullWhen(true)] out string? path)
+        {
+            if (this == default)
+            {
+                FontsThrowHelper.ThrowDefaultInstance();
```
tocsoft, line 132: I've been using a pattern (when using throw helpers) to help inform the compiler that code doesn't continue here is to mark the helper as returning <T> then adding in a return in despite it never really returning... it helps catch unreachable code etc then too..

tocsoft (approved):
Code is fine, however I do vaguely question the need for this feature... it does start to encourage the use of System Fonts in production, as you can always track where you installed fonts from yourself otherwise... the use of system fonts can easily introduce hidden bugs in production code as you deploy into various environments. #179 was a great example of this, because the library author just used system fonts it means that third party devs now couldn't easily influence what was available. (especially bad when using containers as most don't ship with fonts installed at all).

samples/ListFonts/Program.cs
JimBobSquarePants: Yeah, agree, that would have been better. Tbh I used  a struct because I always forget the syntax and it was quicker than searching!

src/SixLabors.Fonts/Font.cs
JimBobSquarePants: Yeah I spotted than and wondered by the runtime didn't just do the same rather than introduce an attribute?

JimBobSquarePants:
Thanks @tocsoft for the review!

The reason I added the feature was because of this request in the runtime https://github.com/dotnet/runtime/issues/49065 I actually got asked about it by @safern when the upcoming changes to System.Drawing support were announced. It seems there is the desire plus, for us it wasn't too difficult. 

There's precedence with DirectWrite also via [`IDWriteFontFace::GetFiles`](https://docs.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwritefontface-getfiles).

tocsoft:
Ok that's fine then... no real harm with the addition then. 👍 

This does make me think making a change to how we expose system fonts might be a good idea... Maybe we should make it so we have a global shared and mutable `FontCollection` that defaults to loading in the standard system fonts (like we have now except mutable), and additionally add an extension over `FontCollection` to add the fonts from the 'standard' system font locations.. That would make sure there is a way to manipulate the global collection to allow a backdoor to resolve issues where fonts are not installed in expected locations or people don't allow using a font directly.

JimBobSquarePants:
Yeah that'd be really useful.

I think... `FontCollectionExtensions.AddSystemFontCollection(this IFontCollection collection)` would probably be my preferred approach for simplicity. We'd simply copy the collection from the `SystemFonts` internal collection.

```c#
var collection = new FontCollection();
collection.AddSystemFontCollection();
```

That's to the point and doesn't require us to manage anything global.

safern:
Thanks for adding this support.

Issue #191, opened 2021-08-11
Helvetica characters are note rendered properly (rendered as rectangles) on macOS


0xced:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp.Drawing
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode (I have only tested with the NuGet package)
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp.Drawing/issues) and [closed](https://github.com/SixLabors/ImageSharp.Drawing/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When rendering text with the Helvetica font on macOS, characters are not rendered properly, they all appear as rectangles.

### Steps to Reproduce

Create the following two files (`RenderHelvetica.csproj` + `Program.cs`) and execute `dotnet run` in the directory in which they were saved. This creates two files, `hello-Arial.png` and `hello-Helvetica.png`.

Arial renders properly:
![hello-Arial](https://user-images.githubusercontent.com/51363/129091702-c4608c0f-5263-43f9-8401-7fe991c196a2.png)

Helvetica fails to render properly (only rectangles are rendered):
![hello-Helvetica](https://user-images.githubusercontent.com/51363/129091744-047381d1-4155-482e-a665-30e49173b60a.png)

RenderHelvetica.csproj
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net5.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="SixLabors.ImageSharp.Drawing" Version="1.0.0-beta13" />
  </ItemGroup>

</Project>
```

Program.cs
```csharp
using SixLabors.Fonts;
using SixLabors.ImageSharp;
using SixLabors.ImageSharp.Drawing.Processing;
using SixLabors.ImageSharp.PixelFormats;
using SixLabors.ImageSharp.Processing;

const int margin = 10;
const int fontSize = 24;
const string text = "Hello World!";

foreach (var fontName in new[] { "Arial", "Helvetica" })
{
    var fontFamily = SystemFonts.Find(fontName);
    var font = new Font(fontFamily, fontSize);
    var rectangle = TextMeasurer.Measure(text, new RendererOptions(font));
[...]

PR #193, opened 2021-08-12
Simplify CMapTable implementation

0xced:
After reading the code of the `CMapTable` class for diagnosing #191 it appears that the code can be improved to more easily understandable.


src/SixLabors.Fonts/Tables/General/CMapTable.cs
```diff
@@ -13,46 +13,29 @@ namespace SixLabors.Fonts.Tables.General
...
-                // keep looking until we have an index thats not the fallback.
-                if (t.TryGetGlyphId(codePoint, out glyphId))
-                {
-                    return true;
-                }
-            }
-
-            // didn't have a windows match just use any and hope for the best
-            foreach (CMapSubTable t in this.Tables)
+            foreach (CMapSubTable t in this.Tables.OrderBy(t => GetPreferredPlatformOrder(t.Platform)))
```
tocsoft, line 36: we will want to do the sort once in the constructor rather than on every glyph lookup.

0xced: Good catch! Addressed in ab083e77a695947c7ec20719ad6c46abb6c3837f.

src/SixLabors.Fonts/Tables/General/CMapTable.cs
```diff
@@ -13,46 +13,29 @@ namespace SixLabors.Fonts.Tables.General
     internal sealed class CMapTable : Table
     {
         internal const string TableName = "cmap";
-        private readonly CMapSubTable[] platformTables;
 
-        public CMapTable(CMapSubTable[] tables)
+        private static readonly Dictionary<PlatformIDs, int> PreferredPlatformOrder = new Dictionary<PlatformIDs, int>
```
JimBobSquarePants, line 17: On my phone but isn’t this already an enum? A simple cast will be much faster than Dictionary + TryGet

0xced: Indeed it's an enum but the `PreferredPlatformOrder` dictionary defines the preferred *order* of the platform ids, i.e. Windows first to preserve the semantics of the previous code.

JimBobSquarePants: Ah yep, I see now, nice!

JimBobSquarePants (approved):
This is much more readable now. Thanks!

PR #195, opened 2021-08-16
Add FontCollection.AddSystemFontCollection

JimBobSquarePants:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/Fonts/pulls) open
- [x] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description

Implements `FontCollection.AddSystemFontCollection()` extension and normalizes the process of adding metrics to the underlying collection.

PR #196, opened 2021-08-16
Better font family not found exception

0xced:
### Prerequisites

### Description
The message now includes all the directory paths where font families were searched and a new `SearchDirectories` property with those same paths.

Before this pull request the exception message was:
> ```
> Caviar Dreams could not be found

After this pull request, the exception message is (specific paths depend on the operating system):
> ```
> The "Caviar Dreams" font family could not be found in the following directories:
>   * /Users/0xced/Library/Fonts/
>   * /Library/Fonts/
>   * /System/Library/Fonts/

Also, the system fonts search paths were improved
* Verbatim string literals are used on Windows
* `%HOME%` is used instead of `~` for Linux and macOS user paths because unlike environment variables, the tilde is not expanded by `Environment.ExpandEnvironmentVariables`
* `/System Folder/Fonts/` was removed on macOS, this was a relic of the Mac OS  9 to OS X transition

### DIFF
--- a/src/SixLabors.Fonts/FontCollection.cs
+++ b/src/SixLabors.Fonts/FontCollection.cs
@@ -16,7 +16,25 @@ namespace SixLabors.Fonts
     /// </summary>
     public sealed class FontCollection : IFontCollection, IFontMetricsCollection
     {
+        private readonly HashSet<string> searchDirectories;
         private readonly HashSet<FontMetrics> metricsCollection = new();
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FontCollection"/> class.
+        /// </summary>
+        public FontCollection()
+            : this(Array.Empty<string>())
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FontCollection"/> class.
+        /// </summary>
+        /// <param name="searchDirectories">The collection of directories used to search for font families.</param>
+        internal FontCollection(IReadOnlyCollection<string> searchDirectories)
+        {
+            this.searchDirectories = new HashSet<string>(searchDirectories ?? throw new ArgumentNullException(nameof(searchDirectories)));
+        }
 
         /// <inheritdoc/>
         public IEnumerable<FontFamily> Families => this.FamiliesByCultureImpl(CultureInfo.InvariantCulture);
@@ -253,7 +271,7 @@ namespace SixLabors.Fonts
                 return family;
             }
 
-            throw new FontFamilyNotFoundException(name);
+            throw new FontFamilyNotFoundException(name, this.searchDirectories);
         }
     }
 }
--- a/src/SixLabors.Fonts/FontCollection.cs
+++ b/src/SixLabors.Fonts/FontCollection.cs
@@ -16,7 +16,7 @@ namespace SixLabors.Fonts
     /// </summary>
     public sealed class FontCollection : IFontCollection, IFontMetricsCollection
     {
-        private readonly HashSet<string> searchDirectories;
+        private readonly HashSet<string> searchDirectories = new();
         private readonly HashSet<FontMetrics> metricsCollection = new();
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


tocsoft:
This message generally only applies in the use case when using 'system fonts' which isn't the pattern that production code should generally be using... (not for any server based workloads anyway where font are a lot more likely to be missing, as they are one of the first things to be striped out of a vm/docker image to reduce size).

This message also doesn't apply when using any form of custom included fonts either by file or stream directly and I feel in that case, where someone has included a local font, such an error message will actually cause more confusion than it would ever save. 

In conclusion there's nothing stopping someone creating a `FontCollection` based on some folders, and then manually adding things from some other source into the collection from elsewhere... what message should be shown then, at the moment you will provide a list of the constructor based folders but that is only half the story and giving potentially incorrect info in a exception would be worse than no additional info at all.

I am however willing to be overruled/convinced otherwise. I think @JimBobSquarePants might also have opinion on this as he has been looking at API polish recently.

JimBobSquarePants:
Yep so #195 adds the ability to add all the system fronts to a collection which would mean this change wouldn't track any of those fonts. If we're tracking the paths searched it should cover all scenarios.

As @tocsoft states, there's other scenarios missed.

For example this code would throw but not include a path.

```c#
var collection  = new FontCollection();

collection.Add("PATH_TO_FONT");

var family = collection.Get("ANOTHER_FONT_NOT_IN_COLLECTION");
```

So I would say the usefulness of the addition in it's current state is limited. 

Let's update the font sources but not implement the exception changes.

0xced:
My initial motivation behind this change is that [I hate vague error messages](https://twitter.com/0xced/status/1455523557455376385). If your error message says that my e-mail is invalid, please tell me _exactly why_ it's considered invalid. If your error message says a font was not found, please tell me _exactly where_ fonts were searched.

Here's what I consider a very good example of such a message (from the [Mongo2Go library](https://github.com/Mongo2Go/Mongo2Go/blob/8aba2bb48ec98b117af031fc633b83d2fb2e427e/src/Mongo2Go/Helper/MongoBinaryLocator.cs#L95-L98)):

```csharp
throw new MonogDbBinariesNotFoundException(
    $"Could not find Mongo binaries using the search patterns \"{_searchPattern}\", \"{Path.Combine(_nugetPrefix, _searchPattern)}\", and \"{Path.Combine(_nugetCachePrefix, _searchPattern)}\".  " +
    $"You can override the search pattern and directory when calling MongoDbRunner.Start.  We have detected the OS as {RuntimeInformation.OSDescription}.\n" +
    $"We walked up to root directory from the following locations.\n {string.Join("\n", searchDirectories)}");
```


> This message also doesn't apply when using any form of custom included fonts either by file or stream directly and I feel in that case, where someone has included a local font, such an error message will actually cause more confusion than it would ever save.

Sorry I haven't property explained this in my initial comment, but the message that includes the searched _directories_ actually only applies to the system fonts collection. (More precisely, font collections that specify their search directories in their constructor). For other font collections, the message will still be **The "Xyz" font family could not be found**, causing no confusion.

> Yep so #195 adds the ability to add all the system fronts to a collection which would mean this change wouldn't track any of those fonts. If we're tracking the paths searched it should cover all scenarios.

Hehe, #195 was opened on the exact same day as this pull request, so I missed it. I have now addressed this scenario in 2ba9bfb118dda235f305a28b954d96ed1506a7eb. A new (well documented) `public IEnumerable<string> SearchDirectories` property has been added on the `IReadOnlyFontCollection` to properly handle the search directories.

Note that if the very recently opened pull request #240 is accepted, it will also have to be adapted to call the new `AddSearchDirectories` method.

> For example this code would throw but not include a path.

```csharp
var collection  = new FontCollection();

collection.Add("PATH_TO_FONT");

var family = collection.Get("ANOTHER_FONT_NOT_IN_COLLECTION");
```

I have addressed this issue by changing the `FontCollection` constructor that takes a collection of search directories fom `internal` to `public`. This now covers the scenario where a consumer of the library scans a custom (non system) directory for additional fonts (87e1c4a1199190aa9a7c3895bec00c3b3f83f9e1).

If no directories are searched and the fonts are added from embedded resources streams for example, then using the parameterless `FontCollection` constructor is the way to go and the exception message won't include any search directories.

> Let's update the font sources but not implement the exception changes.

I think having the list of directories where (system) fonts were searched is a useful addition but If I still haven't convinced you, then please close this pull request and I'll open a new one with only the part that improves the fonts search paths (commit 7603a456bcdf57680d51c361a5e6ad2cd1093b39).

Note: I have rebased my work onto the main branch and force-pushed.

JimBobSquarePants:
@0xced I'll give this all a thorough review as soon as possible. Thanks for your patience. 👍

src/SixLabors.Fonts/FontCollection.cs
```diff
@@ -16,11 +16,37 @@ namespace SixLabors.Fonts
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FontCollection"/> class.
+        /// </summary>
+        /// <param name="searchDirectories">The collection of directories used to search for font families.</param>
+        /// <remarks>
+        /// Use this constructor instead of the parameterless constructor if the fonts added to that collection
+        /// are actually added after searching inside physical file system directories. The message of the
+        /// <see cref="FontFamilyNotFoundException"/> will include the searched directories.
+        /// </remarks>
+        public FontCollection(IReadOnlyCollection<string> searchDirectories)
```
JimBobSquarePants, line 39: Making this constructor public and adding the public property does not make sense to me. We can add Fonts via explicit path and via a stream, neither of which methods are affected by this property.

JimBobSquarePants: Fixed. I split this out to a separate interface.

JimBobSquarePants (approved):
Thanks for your patience on this. I agree that this is a much better result for the user. 👍 

Coverage report is inaccurate as we test all platforms but only report on one.

PR #198, opened 2021-08-26
Bidi + Font Shaping Support

### LINKED ISSUE #66: Support for RTL text

### LINKED ISSUE #150: Ligature's - Add support for ligatures as this is an import feature for correctly rendering Arabic scripts
Add support for ligatures... will help all fonts but they are mandatory for supporting Arabic scripts.

for some reading that might help - https://www.w3.org/TR/alreq/

JimBobSquarePants:
### Prerequisites

### Description

Implementation of bidi layout support + glyph shaping. Includes a complete rewrite of the layout engine.

Information and reference code adapted from
https://docs.microsoft.com/en-us/typography/opentype/spec/
https://github.com/harfbuzz/harfbuzz
https://github.com/foliojs/fontkit
https://github.com/LayoutFarm/Typography

Fixes #150, Fixes #66

Also provides the basic for supporting #8, and #9

#### Implemented features.
- [x] Mixed Bidi layout support LTR, RTL, LTR + RTL
- [x] LayoutMode.Horizontal
- [x] LayoutMode.Vertical
- [x] TextAlign
- [x] CodePoint Script class detection
- [x] Glyph Shaping
- - [x] Default Shaper
- - [x] Arabic Shaper
- - [ ] Indic Shaper (_future_)
- - [ ] Hangul Shaper (_future_)
- - [ ] Hebrew Shaper (_future_)
- - [ ] Khmer Shaper (_future_)
- - [ ] Thai Shaper (_future_)
- - [ ] Myanmar Shaper (_future_)
- - [ ] Universal Shaper (_future_)
- - [x] Gsub Table
- - - [x] Gsub Format 1
- - - [x] Gsub Format 2
- - - [x] Gsub Format 3
- - - [x] Gsub Format 4
- - - [x] Gsub Format 5
- - - [x] Gsub Format 6
- - - [x] Gsub Format 7
- - - [x] Gsub Format 8
- - [x] GPos Table
- - - [x] GPos Format 1
- - - [x] GPos Format 2
- - - [x] GPos Format 3
- - - [x] GPos Format 4
- - - [x] GPos Format 5
- - - [x] GPos Format 6
- - - [x] GPos Format 7
- - - [x] GPos Format 8
- - - [x] GPos Format 9

### Example Output

**Chinese Horizontal**

horizontal-ltr

**Chinese Vertical (_columns are read right to left; the text direction is top to bottom_)**

vertical-rtl

**Combined emoji (_4 scalar values and 7 chars_)**

👩🏽‍🚒a

**Arabic with complex ligatures**

بِسْمِ ٱللَّهِ ٱلرَّحْمَٟنِ ٱلرَّحِيمِ

**Mixed English and Arabic.**

English اَلْعَرَبِيَّةُ English

**Text alignment**

😀A😀-left-bottom
😀A😀-left-center
😀A😀-left-top
😀A😀-center-bottom
😀A😀-center-center
😀A😀-center-top
😀A😀-right-bottom
😀A😀-right-center
😀A😀-right-top

**Word wrapping (_normal_)**

THISIS~1

**Word wrapping (_break all_)**

THISIS~1

**Word wrapping (_keep all_)**

THISIS~1


src/SixLabors.Fonts/RendererOptions.cs
```diff
@@ -173,7 +173,7 @@ internal AppliedFontStyle GetStyle(int index, int length)
 
             return new AppliedFontStyle
             {
-                Start = 0,
+                Start = index,
```
JimBobSquarePants, line 176: Found we were always using zero. Tests were incorrectly expecting that also.

src/SixLabors.Fonts/AppliedFontStyle.cs
```diff
@@ -17,6 +17,8 @@ internal struct AppliedFontStyle
         public int End;
         public bool ApplyKerning;
 
+        // TODO: This should accept all the glyphs in the range to
+        // allow complex substitutions.
         public GlyphMetrics[] GetGlyphLayers(CodePoint codePoint, ColorFontSupport colorFontOptions)
```
JimBobSquarePants, line 22: @tocsoft By my understanding this is our injection point for identifying the correct glyphs for each codepoint. 

It looks like we need to load the entire codepoint buffer though before laying things out in order to remap the glyph ids in context (_we could potentially remove codepoints_).  

I'm seeing though in the run that we load up a style for the entire buffer and never reach the end of that style even though we check per codepoint below. Am I reading that correctly?

tests/SixLabors.Fonts.Tests/GlyphTests.cs
```diff
@@ -137,11 +137,14 @@ public void RenderWoff2Glyphs_EqualToTtfGlyphs()
             var rendererTtf = new ColorGlyphRenderer();
             TextRenderer.RenderTextTo(rendererTtf, testStr, new RendererOptions(fontTtf)
             {
+                // TODO: Check what is different about the woff font.
+                ApplyKerning = false,
```
JimBobSquarePants, line 141: I'm not sure what is happening here. I did a comparison of the rendered output and the results are the same.

@brianpopow you any ideas?

ttf
woff2

brianpopow: @JimBobSquarePants I have just tried this and the test passes. Do i understand this correct, that test fails for you?

JimBobSquarePants: I’ve turned off kerning in the test. Turning it on leads to different vectors in the results for me.

brianpopow: It seems the issue was introduced with d94e0b9e. 

I am not 100% sure yet, but i suspect the change to `AppliedFontStyle.cs`, could be the reason.

JimBobSquarePants: Thanks yeah, massive rewrites to that section. I’ll be the ttf and woff2 fonts doing different things during glyph substitution but I would have thought the gsub table would be identical in both.

JimBobSquarePants:
Hey @tocsoft !! I got combined emojis working

tests/SixLabors.Fonts.Tests/GlyphTests.cs
brianpopow: yeah I think you are right, there is something different going on during the substitution. Not sure how to fix that yet, but this seems to be the reason for the difference.

JimBobSquarePants: Woff isn't doing the substitution. The `I` character is being replaced in the TTF version.

brianpopow: glad you found this

brianpopow: @JimBobSquarePants I would like to change the test a bit and also test for the correct control points count. Is it ok if i commit this change to your branch?

JimBobSquarePants: Yeah. Absolutely!

src/SixLabors.Fonts/Tables/AdvancedTypographic/Gsub/LookupType5SubTable.cs
```diff
@@ -0,0 +1,359 @@
...
+            if (glyphId < 0)
+            {
+                return false;
+            }
+
+            if (this.coverageTable.CoverageIndexOf((ushort)glyphId) > -1)
+            {
+                // TODO: Implement
+                // https://docs.microsoft.com/en-us/typography/opentype/spec/gsub#example-7-contextual-substitution-format-1
+                return false;
```
JimBobSquarePants, line 87: @brianpopow I could do with your considerable spec parsing skills here if you have time. I can follow the spec to determine how to decode the tables however I can't make sense of how to actually implement the substitution. I've been trying t0 understand HarfBuzz but that's hieroglyphs to me.

brianpopow: sure, cant promise I will understand it better, but I will take a look.

src/SixLabors.Fonts/Tables/AdvancedTypographic/CoverageTable.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        private CoverageFormat2Table(CoverageRangeRecord[] records)
+            => this.records = records;
+
+        public override int CoverageIndexOf(ushort glyphId)
+        {
+            for (int i = 0; i < this.records.Length; i++)
+            {
+                CoverageRangeRecord rec = this.records[i];
+                if (rec.StartGlyphId <= glyphId && glyphId <= rec.EndGlyphId)
```
brianpopow, line 94: Is this maybe a mistake here and the comparison should be flipped?

```suggestion
                if (rec.StartGlyphId >= glyphId && glyphId <= rec.EndGlyphId)
```

JimBobSquarePants: Ah no, the glyph has to be within the start and end of the range (inclusive).

JimBobSquarePants:
There's gaps in some of the GSub table support which I could do with help with completing. Lookups 5,6, & 8 which I'm sure someone with an understanding  of C++ or Rust would be able to flesh out.

Complete implementations can be found here.
https://github.com/RazrFalcon/rustybuzz (Rust)
https://github.com/harfbuzz/harfbuzz (C++)

src/SixLabors.Fonts/Tables/AdvancedTypographic/Gsub/LookupType5SubTable.cs
brianpopow: there is alot to chew on here. It will take some time for me to understand that. I am currently trying to understand how harfbuzz does it, hope that will help here.

tests/SixLabors.Fonts.Tests/Tables/AdvancedTypographic/Gsub/GSubTableTests.cs
```diff
@@ -0,0 +1,67 @@
...
+        // TODO: add all other arabic letters
+        [Fact]
+        public void RenderArabicCharacters_WithIsolatedForm_Works()
+        {
+            // arrange
+            Font arabicFont = new FontCollection().Add(TestFonts.ArabicFontFile).CreateFont(8);
+            string testStr = "ب";
+            var rendererTtf = new ColorGlyphRenderer();
+            bool applyKerning = true;
+            Vector2[] expectedControlPoints = new[]
```
tocsoft, line 21: for this test would it not make more sense to figure out what the expected glyph index should be and validate its the one tryign to be rendered rather than looking at control points? 

You can access the rendered glyph index using the `GlyphKeys` collection on  `ColorGlyphRenderer`

brianpopow: yeah thats much better, glad you mentioned it before i did tests for all the other characters 😄

tests/SixLabors.Fonts.Tests/Tables/AdvancedTypographic/Gsub/GSubTableTests.cs
```diff
@@ -60,8 +26,8 @@ public void RenderArabicCharacters_WithIsolatedForm_Works()
             });
 
             // assert
-            Assert.Equal(expectedControlPoints.Length, rendererTtf.ControlPoints.Count);
-            Assert.True(rendererTtf.ControlPoints.SequenceEqual(expectedControlPoints));
+            Assert.Single(rendererTtf.GlyphKeys);
+            Assert.Equal(expectedGlyphIndex, rendererTtf.GlyphKeys[0].GlyphIndex);
```
tocsoft, line 30:
```suggestion
            var glyphKey = Assert.Single(rendererTtf.GlyphKeys);
            Assert.Equal(expectedGlyphIndex, glyphKey.GlyphIndex);
```

makes it slightly easier to read in my opinion

brianpopow: yes, I agree

src/SixLabors.Fonts/Tables/AdvancedTypographic/FeatureTag.cs
```diff
@@ -0,0 +1,820 @@
...
+    /// </summary>
+    internal enum FeatureTag : uint
+    {
+        /// <summary>
+        /// Access All Alternates.
+        /// This feature makes all variations of a selected character accessible. This serves several purposes: An application may not support the feature by which the desired glyph would normally be accessed;
+        /// the user may need a glyph outside the context supported by the normal substitution, or the user may not know what feature produces the desired glyph.
+        /// Since many-to-one substitutions are not covered, ligatures would not appear in this table unless they were variant forms of another ligature.
+        /// </summary>
+        Aalt = 0x61616C74U,
```
JimBobSquarePants, line 18: We actually have a Tag struct that's designed to hold all features, langs, and all sorts.

`Tag.Parse("aalt")` would give you a tag with the `uint` Value.

src/SixLabors.Fonts/Tables/AdvancedTypographic/GSubTable.cs
```diff
@@ -0,0 +1,152 @@
...
+            // TODO: Feature Variations.
+            return new GSubTable(scriptList, featureList, lookupList);
+        }
+
+        public void ApplySubstitions(GlyphSubstitutionCollection collection, ushort index, int count)
+        {
+            collection.GetCodePointAndGlyphIds(index, out CodePoint codePoint, out int _, out IEnumerable<int> _);
+
+            ScriptListTable scriptListTable = this.ScriptList.Default();
+            Tag[] tags = UnicodeScriptTagMap.Instance[CodePoint.GetScript(codePoint)];
```
brianpopow, line 103: I noticed, that the `CodePoint.GetScript` method returns a wrong value for arabic characters. For example "ب", CodePoint 1576 (hex: 628) it returns `CaucasianAlbanian`. My understanding is, that this is because `UnicodeCategory.trie` was build with only latin characters in mind. Is that correct @JimBobSquarePants ?

JimBobSquarePants: Hmmm no, must be a bug in my generation: There should be 1291 Arabic codepoints in there according to the input txt file it’s generated from. Will double check

JimBobSquarePants: Fixed. Generator was fine I was just reading the wrong trie!! 🤣 I've added a few test ranges to confirm it's correct now for now but will flesh out in the future.

JimBobSquarePants:
@brianpopow @tocsoft 

@mhmd-azeez has kindly volunteered to help keep us right with the Arabic shaping output so if you have any questions just ping him.

brianpopow:
Here are some first rendering results for text written from right to left:

farsi

@mhmd-azeez does this look correct to you?

Gillibald:
> Ah…. In the GsubTable we were already able to determine the Script for each individual codepoint. I hoped we’d be able to work like that rather than a single script per text block. How would we mix scripts?

Text needs to be itemized so each script run can be shaped separately. That also comes into play when you enable certain font features for a portion of the text.

mhmd-azeez:
@brianpopow Yes. the correct forms are used and are properly connected. Can you please write it beside an English text to make sure the x-height of the glyphs are correct?

Also here is some Kurdish text to try:

نەما «خانی» هەتا چیڕۆکی عیشقی ئێمە داڕێژێ
دەنا هێشتا لە کوردستان دەژین «زین و مەم»ێکی تر

This also helps test multiline alignment of the text

brianpopow:
> @brianpopow Yes. the correct forms are used and are properly connected. Can you please write it beside an English text to make sure the x-height of the glyphs are correct?

@mhmd-azeez thanks a lot for your review and suggestions, I will give it a try tomorrow.

JimBobSquarePants:
@brianpopow I fixed up those GPos subtables and implemented the update methods. Those tests are still failing though. How did you determine what he correct output should be?

JimBobSquarePants:
>Text needs to be itemized so each script run can be shaped separately. That also comes into play when you enable certain font features for a portion of the text.

@Gillibald Yeah we do that now. Seems to be working well.

https://github.com/SixLabors/Fonts/pull/198/files#diff-db3514b13e65027012cf86809afb38a9b455bc2464a86a4c677dada95a3e5336R211-R241

JimBobSquarePants:
Ah actually @brianpopow , the failing test is related to kerning. We probably shouldn't be applying Kerning and GPOS. Will have to look into how the kerning table works.

JimBobSquarePants:
@mhmd-azeez Could you please post that sample without the line break? I'm having trouble getting it out of the browser in the correct format.  Maybe stick it in a code block?

mhmd-azeez:
@JimBobSquarePants I updated the sample and put it in a code block

JimBobSquarePants:
Whelp! Something's not right!

نەما «خانی» هەتا چیڕۆکی عیشقی ئێمە داڕێژێ-دەنا هێشتا لە کوردستان دەژین «زین و مەم»ێکی تر

Looks like we need to split the text into by lines first then reverse the glyphs within that line.

I don't know what is going on with the brackets though. That's an odd one. If I copy/paste one from the sample into a separate string the direction gets reversed to our output.

mhmd-azeez:
I will try to collect more test cases for you guys to try over the coming days

mhmd-azeez:
Numbers are also a pain because they are actually rendered from left to right!

``
لە ساڵی ١٢٣٤ ڕوویدا
``

image

Fonts:
1. Times new roman
2. Calibri
3. Arabic typesetting
4. Segoe UI

JimBobSquarePants:
We’ll correctly detect the direction for numbers I think and render as expected. 

Brackets are an odd one though. They should ideally be substituted I think? I can’t seem to find relevant info though!

mhmd-azeez:
Yeah, brackets get reversed:

```
 [] {} ()
```
image

And here is some punctuation for good measure:
```
زۆرمان باس کرد؛ باسی ئاو، کەش، هەوای کوردستان.
```
image

JimBobSquarePants:
Hmmm I wonder how that’s handled? We could maintain a list based on the Unicode spec but that seems super hacky.

mhmd-azeez:
This page has a little bit of info on these "mirrored" characters:
- https://www.w3.org/International/articles/inline-bidi-markup/uba-basics#mirroring
- https://www.unicode.org/Public/UCD/latest/ucd/BidiMirroring.txt
- https://unicode.org/reports/tr44/#Bidi_Mirroring_Glyph

brianpopow:
Mixing letters with numbers seems also not correct.

This is the string "لە ساڵی ١٢٣٤ ڕوویدا"

mix_letters_and_characters

mhmd-azeez:
@brianpopow yup. `لە ساڵی` should be on the right and `ڕوویدا` Should be on the left.

JimBobSquarePants:
@brianpopow It's the bidi run order. My offset calc currently only works if the first run in a sequence is LTR. 

I need to have a proper think about this and figure out the correct order to do things. If anyone finds a simple workflow document as to the best approach please share.

src/SixLabors.Fonts/TextLayout2.cs
```diff
@@ -0,0 +1,483 @@
...
+                return Array.Empty<GlyphLayout>();
+            }
+
+            if (options.WrappingWidth > 0)
+            {
+                // Trim trailing white spaces from the text
+                text = text.TrimEnd(null);
+            }
+
+            IEnumerable<TextLine> textLines = ProcessText(text, options);
```
JimBobSquarePants, line 33: @tocsoft @brianpopow This is the beginning of a major rewrite of how we layout text separating shaping, breaking from the actual layout code.

The `IEnumerable<TextLine>` contains a collection `glyph-metrics|bidi-run|codepoint-offset` values which should give us enough information to layout bidirectional horizontal text (_vertical if we want also_).

I still have a few things to figure out but I plan to split layout explicitly for LTR and RTL lines choosing the initial direction based upon the directional value of the first bidi run. There are tricky things to consider though.

https://github.com/SixLabors/Fonts/pull/198#issuecomment-920087412 for example is a real head scratcher. Arabic numbers are actually LTR so there's 3 bidi runs in play.

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -15,415 +18,902 @@ internal class TextLayout
...
+
+                // Reorder them into visual order.
+                orderedRun = LinearReOrder(orderedRun);
+
+                // Now perform a recursive reversal of each run.
+                // From the highest level found in the text to the lowest odd level on each line, including intermediate levels
+                // not actually present in the text, reverse any contiguous sequence of characters that are at that level or higher.
+                // https://unicode.org/reports/tr9/#L2
+
+                // TODO: Fix this. I'm reversing the wrong number of times.
```
JimBobSquarePants, line 728: @brianpopow @tocsoft if one of you could have a look at this for me please that'd be great. I've stuffed up and am reversing the glyphs to incorrect number of times for the mixed numbers/text.

I've been testing the first example using Segoe UI.
https://github.com/SixLabors/Fonts/pull/198#issuecomment-919887482

As you can see the glyph runs are in the correct sequence, it's just the actual glyph ordering.
لە ساڵی ١٢٣٤ ڕوویدا

brianpopow: I think it looks ok now:

bidi_test

brianpopow:
> Another test case to check: Mathematical equations
> 
> ```
> ١١س + ٨ج = ٨٥١
> ```
> 
> image
> 
> Font: Noto Naskh Arabic

@mhmd-azeez thanks for the suggestion. Here is how it looks:
math_formula

The orientation of the characters look correct, but the plus and equal signs are not drawn as a placeholder character. I need to investigate why, not sure yet.

tocsoft:
> The orientation of the characters look correct, but the plus and equal signs are not drawn as a placeholder character. I need to investigate why, not sure yet.

I would guess the glyphs are just not in the font, and the render above is using a fallback font for those 2 characters.

mhmd-azeez:
@tocsoft you're right! I looked at the font using https://fontdrop.info/ and it doesn't have `+` and `=` symbols. I was using LibreOfficer Writer and it seems like it just falls back to another font

brianpopow:
> That's an interesting point though, does the library support font fallbacks?
yes, we do.

I have tried a different font called nazli, which has + and = signs: 

math_nazli


Looks correct @mhmd-azeez , right?

mhmd-azeez:
@brianpopow yes, it is correct :+1:

JimBobSquarePants:
@brianpopow I've just pushed changes so that GPOS now filters for features. I _think_ it's correct but I was going off the FontKit implementation rather than HarfBuzz. I could do with a review of the changes just to be sure. 

I did find one oddity in the original implementation. hashsets were being assigned at offsets and queried at indexes which was working since we're not skipping codepoints but potentially they can be skipped which meant I had to move things around a little. This added some overhead to the lookups which I'd love to figure out a way to avoid since a lookup could now fail without them. I may have overengineered things though!

brianpopow:
> @brianpopow I've just pushed changes so that GPOS now filters for features. I _think_ it's correct but I was going off the FontKit implementation rather than HarfBuzz. I could do with a review of the changes just to be sure.
> 
> I did find one oddity in the original implementation. hashsets were being assigned at offsets and queried at indexes which was working since we're not skipping codepoints but potentially they can be skipped which meant I had to move things around a little. This added some overhead to the lookups which I'd love to figure out a way to avoid since a lookup could now fail without them. I may have overengineered things though!

I am currently working on tests for GSUB, there are still a few things which are not working as expected. See failing tests. I want to figure this out before i move on to GPOS. I will review the changes to GPOS when im done with this.

brianpopow:
> > @brianpopow I've just pushed changes so that GPOS now filters for features. I _think_ it's correct but I was going off the FontKit implementation rather than HarfBuzz. I could do with a review of the changes just to be sure.

@JimBobSquarePants commit 39ab42b looks good to me, cant spot anything obvious which should be changed.

I will continue with gathering test fonts for GPOS, so we can test if everything works out as intended. One thing i spotted so far:
While LookupType1SubTable, Format 1 does apply the offset, the rendered output looks cut off as if the bounding box is not moved:


Cant spot the issue so far. The offset is applied to the bounding box.

Font is: `GposLookupType1Format1.ttf`
Teststring: `\u0015\u0014`

JimBobSquarePants:
Ok. I'll have a look. I need to gather some more info for the sub tables also that we don't have just now. Whether the Codepoint is a mark or base glyph. I should be able to add it using existing category data. We need it for GSub Format 4 also as we're substituting things we shouldn't.

JimBobSquarePants:
@brianpopow is this what you expect?

brianpopow:
> 
> @brianpopow is this what you expect?

That looks much better! The 20 should be placed 200 units to the left. Its a bit hard to tell if that's the case. Maybe adding some more characters to the right could help visually.

JimBobSquarePants:
Yeah I'd say this is correct. 

Original


Shifted
-- - Copy

JimBobSquarePants:
@brianpopow Found an interesting issue with Gsub Format4 subtable - Ligature substitution. 

If I test the following string `first\n\n\n\nl` with Segoe UI of Open Type then the `i` following the `f` is removed.

I had a look at Harfbuzz to see if there was something simple I could implement but unfortunately I cannot make sense of it.
https://github.com/harfbuzz/harfbuzz/blob/9247d24d7f41f820d7ea79089b482410b927edc8/src/hb-ot-layout-gsubgpos.hh#L931

JimBobSquarePants:
Sorry @brianpopow my bad... It was Segoe UI with the OpenSans-Regular.ttf as a fallback. 

I've updated the code to prevent look for fallbacks among control characters which fixes that issue however.... If you try to write that text using OpenSans the substitution takes place and it actually crashes the drawing processor with an IndexOutOfRangeException!

JimBobSquarePants:
@brianpopow just did a big refactor to pass the current feature as the input sequence matcher should check against it. Unfortunately the Open Sans issue still exists but the API is much easier to work with now. Big speedup also. About 2X faster for a test run on the Ubuntu CI 6secs vs 12.

brianpopow:
> @brianpopow just did a big refactor to pass the current feature as the input sequence matcher should check against it. Unfortunately the Open Sans issue still exists but the API is much easier to work with now. Big speedup also. About 2X faster for a test run on the Ubuntu CI 6secs vs 12.

I will try to figure out the issue with Open Sans.

brianpopow:
> Unfortunately the Open Sans issue still exists 

@JimBobSquarePants can you explain how exactly you can trigger the issue with OpenSans? I tried this test string `first\n\n\n\nl`
with segeoui font and OpenSans-Regular as the fall back font. This does not trigger an exception.

JimBobSquarePants:
@brianpopow stick this code in the Program.cs file in the samples project.

```c#
RenderText(font2, "first\n\n\n\nl", pointSize: 20);
```

JimBobSquarePants:
I'm relegated currently to attaching a debugger to https://fontkit-demo.vercel.app/ and loading the font+text as I cannot get a local build of it to run. So far I can see a substitution is taking place with the same return. However I have no idea what happens next.

JimBobSquarePants:
@brianpopow I'm 100% confident now that the issue is with ImageSharp.Drawing.

FontKit returns exactly the same glyph which has Id 564. In the font this represents codepoint 0xFB01 which is a Latin small ligature `fi` 

image

If I directly reference my local copy of ImageSharp.Drawing from the sample program in the Fonts solution I can attach the debugger. The exception is thrown in `ActiveEdgeList.LeaveEdge(idx)`. If I disable the exception I get the following output.

ﬁ

As you can see the bounding box is correct and the `f` portion is renderable before the issue occurs. I think we should move on and try to fix the issue in Drawing once this is merged. @antonfirsov I'll absolutely need your help then. The rasterizer is a black box for me.

JimBobSquarePants:
Bugger.... Wind that all back. I just created a test project using latest nightlies and I can render it!
fi

Now to compare the font metrics....

JimBobSquarePants:
I'm thoroughly confused now. I can get it to throw using the latest Fonts master but I simply cannot get it to throw using the latest from MyGet.

JimBobSquarePants:
It's the woff version that's the issue! Likely a bad conversion. If you load the woff version followed by the ttf version into the font collection it cannot tell the difference and ignores the ttf. Not sure what to do about that behaviour, better equality comparison I think.

ﬁ

brianpopow:
> It's the woff version that's the issue! Likely a bad conversion. If you load the woff version followed by the ttf version into the font collection it cannot tell the difference and ignores the ttf. Not sure what to do about that behaviour, better equality comparison I think.
> 
> ﬁ

It has something to do with the substitution also. If you disable substitutions the error does not happen.

JimBobSquarePants:
>It has something to do with the substitution also. If you disable substitutions the error does not happen. 

Yep so the substitution replaces the individual `f` & `i` with the single `fi` glyph which in the woff font is buggy (_The substitution is correct though_). You can trigger the same exception with that font by simply trying to layout the single `fi` glyph without kerning enabled.

JimBobSquarePants:
@brianpopow I've gotten a little side tracked and am just doing a little bit of testing and fixing of the woff2 loader. 

I've substituted the Opens Sans font I have locally for the [v3 version](https://google-webfonts-helper.herokuapp.com/fonts/open-sans?subsets=cyrillic,cyrillic-ext,greek,greek-ext,hebrew,latin,latin-ext,vietnamese) so I could get the proper woff2 file rather than a generated one. I couldn't load the file at first as the header reader was telling me there were duplicates but after a little refactoring of the table loading code based on the [official source](https://github.com/google/woff2/blob/master/src/woff2_dec.cc) I was able to get it loading. FontCreator can load and render the glyph well enough but we only seem to be detecting half the control points (_49 vs 87 where all points are equal until the cutoff_) when compared to the ttf version. I suspect something is incomplete. I've encountered that a fair amount in the LayoutFarm source.

`\uFB01` TTF
ﬁ-ttf

`\uFB01` WOFF2
ﬁ-woff2

Incidentally the woff font cannot be loaded by us at all. It complains about a backwards seek (_FontCreator opens it just fine_).

brianpopow:
> @brianpopow I've gotten a little side tracked and am just doing a little bit of testing and fixing of the woff2 loader.
> 
> I've substituted the Opens Sans font I have locally for the [v3 version](https://google-webfonts-helper.herokuapp.com/fonts/open-sans?subsets=cyrillic,cyrillic-ext,greek,greek-ext,hebrew,latin,latin-ext,vietnamese) so I could get the proper woff2 file rather than a generated one. I couldn't load the file at first as the header reader was telling me there were duplicates but after a little refactoring of the table loading code based on the [official source](https://github.com/google/woff2/blob/master/src/woff2_dec.cc) I was able to get it loading. 
> ...

@JimBobSquarePants I would suggest to create a separate branch to fix this woff loading issues. This PR is already quite large.
Maybe we can start of with a unittest, which reproduces the issue(s), something like `RenderWoff2Glyphs_EqualToTtfGlyphs`.

JimBobSquarePants:
>I would suggest to create a separate branch to fix this woff loading issues. 

@brianpopow Yep, completely agree.

brianpopow:
> I could get the proper woff2 file rather than a generated one. I couldn't load the file at first as the header reader was telling me there were duplicates but after a little refactoring of the table loading code based on the

@JimBobSquarePants how did you manage to get the woff2 file to load. When I skip the duplicate table entry, I get a brotli decompression error.

JimBobSquarePants:
@brianpopow [See here](https://github.com/SixLabors/Fonts/pull/209/files#diff-3ef827409aef86b4ef7d3dc976ca9d4cb912e412b2b69d9688adfe3b73a42a74R48) I just copied the way Google do it.

JimBobSquarePants:
@brianpopow Just updated GPOS LookupType4SubTable to what I think is the correct approach. I couldn't find a test though and I wasn't sure where you were getting the expected info from.

brianpopow:
> @brianpopow Just updated GPOS LookupType4SubTable to what I think is the correct approach. I couldn't find a test though and I wasn't sure where you were getting the expected info from.

@JimBobSquarePants there is no test yet. Usually i try to get it to look like harfbuzz and then write a test comparing the glyph rectangles to expected rectangles. 
There is a test font `GposLookupType4Format1.ttf` from harfbuzz.
Use the string `\u0012\u0013`

This should look like:
1819_aftergpos

Actually looks like:
output

My understanding is, we need to move the glyph at `index` (the 19) to the baseGlyph position (the "not mark" glyph) and the apply the offsets. I could not figure out how to properly achieve that. Maybe we need a better example font. I think some arabic scripts use this feature to add symbols over character to describe how vowels are pronounced (they called **fatheh, kaßreh and sammeh**). Any chance @mhmd-azeez may know a font which uses the gpos [Mark-to-Base Attachment Positioning](https://docs.microsoft.com/en-us/typography/opentype/spec/gpos#lookup-type-4-mark-to-base-attachment-positioning-subtable) for that?

JimBobSquarePants:
Ah ok… So that’s the output after my changes?

brianpopow:
> Ah ok… So that’s the output after my changes?

yes it is

JimBobSquarePants:
Cool. If you don’t beat me to it I’ll have a look again tomorrow

JimBobSquarePants:
@brianpopow there you go. The original doesn't seem to account for vertical layout so I took that into consideration also.



Btw, dunno if you've seen this but it's damn useful
https://fontkit-demo.vercel.app/

brianpopow:
> @brianpopow there you go. The original doesn't seem to account for vertical layout so I took that into consideration also.
> 

Good job, that looks correct! I will add a test for that.

> Btw, dunno if you've seen this but it's damn useful https://fontkit-demo.vercel.app/
> 

I did not know that, looks very useful, thanks.

mhmd-azeez:
@brianpopow This is beyond my area of expertise so I am not sure. As you explained these diacritics are used to show how the word is supposed to be pronounced. Most modern Arabic texts don't have them though because they are too time consuming to add and native Arabic speakers can pronounce the text without needing the diacritics usually. Here is an example sentence in Arabic that uses [Fatheh](https://www.compart.com/en/unicode/U+064E), [Kasrah](https://www.compart.com/en/unicode/U+061A), [Sukun](https://www.compart.com/en/unicode/U+0652), [Kasratan](https://www.compart.com/en/unicode/U+064D), and [Shaddah + Damma](https://www.compart.com/en/unicode/U+FC61):

```
كَيْفَ تُمِدُّ جِسْمَكَ بِالْـماءِ بِطَريقَةٍ صَحِيحَةٍ؟
```

For comparison, here is the same sentence without the diacritics:

```
كيف تمد جسمك بالماء بطريقة صحيحة؟
```

Times New Roman:
image

Calibri:
image

Arabic Typesetting:
image

[Othmani Font](http://legionfonts.com/fonts/othmani):
image

Note: There are other diacritics used in Arabic:
https://en.wikipedia.org/wiki/Arabic_diacritics

brianpopow:
@mhmd-azeez thanks for the examples, those are helpful to find a font with this feature!

brianpopow:
@JimBobSquarePants there is still a bug in loading the gpos tables, i guess its related to Lookup Type 5. Will look into it tomorrow.

src/SixLabors.Fonts/Tables/AdvancedTypographic/GPos/LookupType6SubTable.cs
```diff
@@ -0,0 +1,132 @@
...
+                    return false;
+                }
+
+                // TODO: fontkit checks here, if Marks belonging to the same base or Marks belonging to the same ligature component.
+                int prevIdx = index - 1;
+                ushort prevGlyphId = collection[prevIdx][0];
+                GlyphShapingData prevData = collection.GetGlyphShapingData(prevIdx);
+                if (!CodePoint.IsMark(prevData.CodePoint))
+                {
+                    return false;
```
brianpopow, line 115: `\0012` should be a Mark in the font GposLookupType6Format1.ttf, but `CodePoint.IsMark` returns false here. Maybe checking if its a mark via CodePoint is not correct here.

JimBobSquarePants: Cool. I’ll have a look tonight for you

JimBobSquarePants: Looking now.

brianpopow: if you load the font with Font Creator, you can see the `OT Class` of `\0012` to be Mark.

JimBobSquarePants: Ok. So our `CodePoint.IsMark()` check is correct, however this is not what FontKit nor HarfBuzz are checking. They both maintain `LigatureId` and `LigatureComponent` properties for each glyph and check those. 

We need to add those properties to the `GlyphShapingData` type and also a `SetGlyphShapingData(index, data)` method to `GlyphPositioningCollection`. We'll have to ensure we look back through the FontKit source to see when those properties are ever set.

JimBobSquarePants: We very likely need GDEF table support.
https://docs.microsoft.com/en-us/typography/opentype/spec/gdef

JimBobSquarePants: Hold on..... Yep. It's a combining mark when you check its LineBreakClass..... I need to do some research to make the IsMark test more robust!

JimBobSquarePants: Ok So I did a whole bunch of research and augmented the method with the following code to cover all available ranges however that then broke lookups 4 and 5.

https://en.wikipedia.org/wiki/Combining_character
https://unicodesymbols.fandom.com/wiki/Category:Combining_Mark_Set

```c#
public static bool IsMark(CodePoint codePoint)
{
    // https://en.wikipedia.org/wiki/Combining_character

    // Combining Diacritical Marks
    if (UnicodeUtility.IsInRangeInclusive(codePoint.value, 0x0300, 0x036F))
    {
        return true;
    }

    // Combining Diacritical Marks Extended
    if (UnicodeUtility.IsInRangeInclusive(codePoint.value, 0x1AB0, 0x1AFF))
    {
        return true;
    }

    // Combining Diacritical Marks Supplement
    if (UnicodeUtility.IsInRangeInclusive(codePoint.value, 0x1DC0, 0x1DFF))
    {
        return true;
    }

    // Combining Diacritical Marks for Symbols
    if (UnicodeUtility.IsInRangeInclusive(codePoint.value, 0x20D0, 0x20FF))
    {
        return true;
    }

    // Combining Half Marks
    if (UnicodeUtility.IsInRangeInclusive(codePoint.value, 0xFE20, 0xFE2F))
    {
        return true;
    }

    return IsCategoryMark(GetGeneralCategory(codePoint)) || GetLineBreakClass(codePoint) == LineBreakClass.CM;
}
```

I'm not sure what the next step is. Do we simply check the linebreak class in lookup 6?

brianpopow: My understanding so far was, that we need to implement the GDEF table and get the info, if the glyph is a mark from there, right?

JimBobSquarePants: Unfortunately so. Unicode support should be enough but it doesn’t seem to be.

JimBobSquarePants: Fontkit cheats by storing multiple codepoints against a glyph following a removal and counting that as a mark. Theoretically we could do the same but it’s yet another array

brianpopow: Ok, I will start then implementing the GDEF table

brianpopow: so far it seems to work out as planed: I use now the GDEF table if present to get the glyph class otherwise CodePoint.IsMark in LookupType6SubTable

JimBobSquarePants: Awesome! That was quick!

src/SixLabors.Fonts/Tables/AdvancedTypographic/GPos/LookupType5SubTable.cs
```diff
@@ -0,0 +1,151 @@
...
+                if (ligatureIndex < 0)
+                {
+                    return false;
+                }
+
+                LigatureAttachTable ligatureAttach = this.ligatureArrayTable.LigatureAttachTables[ligatureIndex];
+                ushort markGlyphId = glyphId;
+                ushort ligGlyphId = baseGlyphId;
+
+                // TODO: figure out how to calculate the compIndex, see fontKit.
```
brianpopow, line 140: not sure howto fix this TODO here. Im not sure what the ligatureID is supposed to be. @JimBobSquarePants maybe you have an idea?

to quote from harfbuzz
```
We must now check whether the ligature ID of the current mark glyph
is identical to the ligature ID of the found ligature.  If yes, we
can directly use the component index.  If not, we attach the mark
glyph to the last component of the ligature.
```

Its `hb-ot-layout-gpos-table.hh` line 1202

JimBobSquarePants: So, looking at FontKit the ligature Id would be set in a single places. GSUB  `LookupType4SubTable`

The default value (taken from the processor, the `GlyphSubstitutionCollection` in our instance) would be `1`.

image

They seem to carry a lot of info around with the data right through the process.

image

JimBobSquarePants:
@brianpopow I've just pushed the font from the issue mentioned in the ligature substitution table. https://bugzilla.gnome.org/show_bug.cgi?id=437633

I got it from here.
https://sourceforge.net/projects/al-anvar/

Compared to FontKit it looks like we've still got some work to do, Seems like a whole set of the glyphs are missing.
image

JimBobSquarePants:
I'll have a look tomorrow. Late here now. I've been doing some more testing of the ligature substitution table and there definitely seems to be nothing taking place there... (_real headscratcher that one_)

JimBobSquarePants:
@brianpopow I've written the basic functionality for a `SkippingGlyphIterator` type which is important for determining which glyphs to operate upon and match. Currently it's a bit awkward to implement as all our individual matching functions should be using it but they all operate independently. I suggest we implement a `Match` method in `AdvancedTypographicUtils` that accepts a `Func` that can be called from all our matching methods. FontKit and HarfBuzz to the same.

JimBobSquarePants:
@brianpopow I managed to get the GSUB feature application working with the iterator by refactoring to do the following. We still preserve the correct application order plus can now skip when required. However, this does not change the result.

```c#
foreach (Tag stageFeature in shaper.GetShapingStageFeatures())
{
    List<(Tag Feature, ushort Index, LookupTable LookupTable)> lookups = this.GetFeatureLookups(stageFeature, current);
    if (lookups.Count == 0)
    {
        continue;
    }

    int currentCount = collection.Count;
    SkippingGlyphIterator iterator = new(fontMetrics, collection, index, default);

    // Apply features in order.
    foreach ((Tag Feature, ushort Index, LookupTable LookupTable) featureLookup in lookups)
    {
        Tag feature = featureLookup.Feature;
        iterator.Reset(index, featureLookup.LookupTable.LookupFlags);

        while (iterator.Index < index + count)
        {
            List<TagEntry> glyphFeatures = collection.GetGlyphShapingData(iterator.Index).Features;

            // TODO: No LINQ.
            if (!glyphFeatures.Any(x => x.Tag == featureLookup.Feature && x.Enabled))
            {
                iterator.Next();
                continue;
            }

            featureLookup.LookupTable.TrySubstitution(fontMetrics, this, collection, featureLookup.Feature, iterator.Index, count - (iterator.Index - index));
            iterator.Next();

            // Account for substitutions changing the length of the collection.
            if (collection.Count != currentCount)
            {
                count = (ushort)(count - (currentCount - collection.Count));
                currentCount = collection.Count;
            }
        }
    }
}
```

Gillibald:
`Chinese Vertical (columns are read right to left; text direction is left to right)` -> `Chinese Vertical (columns are read right to left; text direction is top to bottom)`

JimBobSquarePants:
Thanks @Gillibald just a bad copy/paste will fix. We’re definitely doing the right thing

Gillibald:
You did a really great job here. So impressive. Hopefully, this will be adopted by many projects so this can improve even further. .Net is used everywhere so there should be first-class text processing without any extra native dependencies. Your work is the first starting point to make that possible.

antonfirsov (commented):
A bunch of random notes on the code without understanding what's happening.

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -13,88 +16,529 @@ namespace SixLabors.Fonts
...
     {
-        internal static TextLayout Default { get; set; } = new TextLayout();
-
-        /// <summary>
-        /// Generates the layout.
-        /// </summary>
-        /// <param name="text">The text.</param>
-        /// <param name="options">The style.</param>
-        /// <returns>A collection of layout that describe all that's needed to measure or render a series of glyphs.</returns>
+        internal static TextLayout Default { get; set; } = new();
```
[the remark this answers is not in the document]
JimBobSquarePants: It's an odd one. `TextMeasurer` accepts the instance as a parameter and stores it as a private field. Not sure whether @tocsoft had plans to allow replacing parts.

src/SixLabors.Fonts/RendererOptions.cs
```diff
@@ -82,34 +81,25 @@ public RendererOptions(Font font, float dpiX, float dpiY, Vector2 origin)
...
+        /// Gets or sets the width of the tab. Measured as the distance in spaces.
         /// </summary>
-        /// <value>
-        /// The width of the tab.
-        /// </value>
         public float TabWidth { get; set; } = 4;
 
         /// <summary>
-        /// Gets or sets a value indicating whether [apply kerning].
+        /// Gets or sets a value indicating whether to apply kerning.
```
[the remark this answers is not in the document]
JimBobSquarePants: So I'm currently doing this wrong.

There's two types of kerning in an OpenType font. Historical True Type kerning (_kern_) tables and modern glyph positioning (_gpos_) tables.

We're actually turning off glyph positioning and substitution with this switch when we shouldn't be.  

We should probably be allowing GSUB and GPOS always and only toggling kerning in the kern table. The [spec](https://www.w3.org/TR/css-fonts-3/#font-kerning-prop)states the same.

There's a CSS option for turning off [common ligatures](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-ligatures) which we should probably adopt also.

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -105,325 +549,560 @@ public IReadOnlyList<GlyphLayout> GenerateLayout(ReadOnlySpan<char> text, Render
...
+            return new TextBox(textLines, scaledMaxAscender, scaledMaxDescender, scaledMaxLineHeight, scaledMaxLeftSideBearing);
+        }
 
-                        if (finalWidth < glyphWidth)
-                        {
-                            // If we are not going to tab at least a glyph width add another tabstop
-                            // to it ??? TODO: Should I be doing this?
-                            finalWidth += tabStop;
-                        }
+        internal sealed class TextBox
```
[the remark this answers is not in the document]
JimBobSquarePants: It's a container for a collection of text lines, so just once per render.

src/SixLabors.Fonts/Unicode/BidiRun.cs
```diff
@@ -1,30 +1,38 @@
 // Copyright (c) Six Labors.
 // Licensed under the Apache License, Version 2.0.
 
+using System;
 using System.Collections.Generic;
 
 namespace SixLabors.Fonts.Unicode
 {
-    internal readonly struct BidiRun
+    internal readonly struct BidiRun : IEquatable<BidiRun>
```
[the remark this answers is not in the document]
JimBobSquarePants: Most of the heavy documentation was for my purpose. I've found few things as complicated as GSUB/GPOS.

src/SixLabors.Fonts/RendererOptions.cs
```diff
@@ -134,58 +124,41 @@ public RendererOptions(Font font, float dpiX, float dpiY, Vector2 origin)
...
+        /// Gets or sets the vertical alignment of the text box.
         /// </summary>
         public VerticalAlignment VerticalAlignment { get; set; }
 
         /// <summary>
-        /// Gets or sets the rendering origin.
+        /// Gets or sets the layout mode for the text lines.
         /// </summary>
-        public Vector2 Origin { get; set; } = Vector2.Zero;
+        public LayoutMode LayoutMode { get; set; }
```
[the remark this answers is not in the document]
JimBobSquarePants: Good question....

You would never actually render an Arabic font vertically, the language is explicitly horizontal and right-to-left so you would always rotate the glyphs and render them sideways (_we don't do that automatically_).

The visual layout (order) of glyphs is determined by the bidirectional rules of each script detected from the codepoints contained within the input string. For vertical alignment there is some default substitution of known codepoints but then shaping (gsub/gpos) rules apply. Our implementation of these follows HarfBuzz which work well. Glyphs that should be positioned together are tightly bound and vertical layout would not break that.

`LayoutMode.VerticalLeftRight`, and `LayoutMode.VerticalRightLeft` are really an opt-in thing for users who want to render languages vertically that support them. 

`LayoutMode.HorizontalTopBottom` was just showing off 😉

src/SixLabors.Fonts/Buffer{T}.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace SixLabors.Fonts
+{
+    /// <summary>
+    /// An disposable buffer that is backed by an array pool.
+    /// </summary>
+    /// <typeparam name="T">The type of buffer element.</typeparam>
+    internal struct Buffer<T> : IDisposable
```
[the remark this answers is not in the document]
JimBobSquarePants: Yep agreed. Need to remove the interface though and use compiler pattern matching for ref structs. Done

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -105,325 +549,560 @@ public IReadOnlyList<GlyphLayout> GenerateLayout(ReadOnlySpan<char> text, Render
...
+            public float ScaledMaxLineGap { get; }
+
+            public float ScaledMaxLeftSideBearing { get; }
+
+            public ReadOnlyCollection<TextLine> TextLines { get; }
+
+            public TextDirection TextDirection() => this.TextLines[0][0].TextDirection;
+        }
+
+        internal sealed class TextLine
```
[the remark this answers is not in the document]
JimBobSquarePants: Depends on the number of lines rendered. Usually very small numbers.

samples/DrawWithImageSharp/Program.cs
```diff
@@ -36,6 +36,26 @@ public static void Main(string[] args)
...
             FontFamily emojiFont = SystemFonts.Get("Segoe UI Emoji");
             FontFamily uiFont = SystemFonts.Get("Segoe UI");
+            FontFamily arabicFont = SystemFonts.Get("Dubai");
+
+            FontFamily bugzilla = fonts.Add(@"Fonts\me_quran_volt_newmet.ttf");
+
+            //RenderText(uiFont, "Soft\u00ADHyphen", pointSize: 72);
+            RenderText(bugzilla, "بِسْمِ ٱللَّهِ ٱلرَّحْمَٟنِ ٱلرَّحِيمِ", pointSize: 72);
+            return;
+            // TODO: Test and fix tomorrow. Gsub Lookup 4.
```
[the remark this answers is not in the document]
JimBobSquarePants: Yeah. I'll likely delete most of it during a cleanup though.

src/SixLabors.Fonts/Tables/AdvancedTypographic/UnicodeScriptTagMap.cs
```diff
@@ -0,0 +1,192 @@
...
+using System.Collections.Generic;
+using SixLabors.Fonts.Unicode;
+
+namespace SixLabors.Fonts.Tables.AdvancedTypographic
+{
+    /// <summary>
+    /// Provides a map from Unicode <see cref="ScriptClass"/> to OTF <see cref="Tag"/>.
+    /// <see href="https://docs.microsoft.com/en-us/typography/opentype/spec/scripttags"/>
+    /// </summary>
+    internal sealed class UnicodeScriptTagMap : Dictionary<ScriptClass, Tag[]>
```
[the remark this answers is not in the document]
JimBobSquarePants: We use `TryGet` I believe so not currently.

src/SixLabors.Fonts/Tables/AdvancedTypographic/TagEntry.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System.Diagnostics;
+
+namespace SixLabors.Fonts.Tables.AdvancedTypographic
+{
+    [DebuggerDisplay("Tag: {Tag}, Enabled: {Enabled}")]
+    internal class TagEntry
```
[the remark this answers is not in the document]
JimBobSquarePants: I'll check.

src/SixLabors.Fonts/GlyphSubstitutionCollection.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// </summary>
+        public int Count => this.offsets.Count;
+
+        /// <summary>
+        /// Gets or sets the running id of any ligature glyphs contained withing this collection are a member of.
+        /// </summary>
+        public int LigatureId { get; set; } = 1;
+
+        /// <inheritdoc />
+        public ReadOnlySpan<ushort> this[int index] => this.glyphs[this.offsets[index]].GlyphIds;
```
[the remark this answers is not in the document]
JimBobSquarePants: So....

A Glyph Id is an identifier for a glyph in a font. Previously Glyph Index and Glyph Id were used interchangeably but I believe that to be incorrect. Lookups for glyphs are performed against a given codepoint and there's no rule that states they should be stored sequentially (_Only that zero is always the fallback glyph_). 

There's more than one at a given index within the collection because the collection is actually more representative of a jagged array (_though we don't use that for API ease of use_). You can have more than one glyph representing a single codepoint following substitution (_e.g. emojis_)

src/SixLabors.Fonts/Tables/AdvancedTypographic/AttachPoint.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.Fonts.Tables.AdvancedTypographic
+{
+    internal struct AttachPoint
+    {
+        public ushort[] PointIndices;
```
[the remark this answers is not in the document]
JimBobSquarePants: Oooh! I'll check this out!

src/SixLabors.Fonts/Tables/AdvancedTypographic/GPos/LigatureAttachTable.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace SixLabors.Fonts.Tables.AdvancedTypographic.GPos
+{
+    /// <summary>
+    /// Each LigatureAttach table consists of an array (componentRecords) and count (componentCount) of the component glyphs in a ligature.
+    /// The array stores the ComponentRecords in the same order as the components in the ligature.
+    /// The order of the records also corresponds to the writing direction — that is, the logical direction — of the text.
+    /// For text written left to right, the first component is on the left; for text written right to left, the first component is on the right.
+    /// </summary>
+    internal class LigatureAttachTable
```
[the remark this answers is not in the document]
JimBobSquarePants: Yeah, I'll have a look.

src/SixLabors.Fonts/Tables/AdvancedTypographic/CoverageTable.cs
```diff
@@ -0,0 +1,148 @@
...
+    /// The GSUB, GPOS, and GDEF tables rely on this notion of coverage.
+    /// If a glyph does not appear in a Coverage table, the client can skip that subtable and move
+    /// immediately to the next subtable.
+    /// <see href="https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#coverage-table"/>
+    /// </summary>
+    internal abstract class CoverageTable
+    {
+        public abstract int CoverageIndexOf(ushort glyphId);
+
+        public abstract IEnumerator<ushort> GetEnumerator();
```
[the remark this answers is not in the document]
JimBobSquarePants: That's a hangover of reading the Typography source. It can be removed.

src/SixLabors.Fonts/RendererOptions.cs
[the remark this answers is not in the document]
antonfirsov: Hm, I only understand half of this, but if we are not sure the property controls the right thing, maybe better to remove it from the public API for now, and open a tracking issue to figure it out later?

src/SixLabors.Fonts/Tables/AdvancedTypographic/UnicodeScriptTagMap.cs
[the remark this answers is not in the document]
antonfirsov: It's easy to implement `TryGet`: We can store `null` in the array for values which are not in `UnicodeScriptTagMap.Instance` but for some reason defined in `ScriptClass`.

Also note that all current values are arrays of `Length == 1`, so maybe we can store `Tag?` instead of `Tag[]`?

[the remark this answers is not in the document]
antonfirsov: Note that I don't expect all these nits to be resolved in the PR, but would be nice to add TODO notes or a tracking issue documenting the code quality / perf leftovers.

src/SixLabors.Fonts/RendererOptions.cs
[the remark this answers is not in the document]
JimBobSquarePants: Yep. I've done some research and know how this works now. I'll push a fix in a day.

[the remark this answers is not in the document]
JimBobSquarePants: Pushed. We now match the CSS specification with a new `KerningMode` property on `RendererOptions`. Unfortunately this will cause runtime errors in ImageSharp.Drawing until we update the `TextOptions` to reflect this (due to copying from ROps to TOps to new ROps instance)

src/SixLabors.Fonts/Tables/AdvancedTypographic/CoverageTable.cs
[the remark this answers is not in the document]
JimBobSquarePants: Done

src/SixLabors.Fonts/Tables/AdvancedTypographic/UnicodeScriptTagMap.cs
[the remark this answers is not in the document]
JimBobSquarePants: I recall why this is an array now. Historically there's been changes to features over the years with some made obsolete that we still need to associate with a script tag. There's evidence of various substitutions scattered throughout the HarfBuzz code but I haven't gotten around to collating them.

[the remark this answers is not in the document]
JimBobSquarePants: Turns out I already had a TODO for that in the code! 😄

src/SixLabors.Fonts/Tables/AdvancedTypographic/AttachPoint.cs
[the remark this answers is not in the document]
JimBobSquarePants: Ah yeah, this has something to do with font variations which we don't support yet.

src/SixLabors.Fonts/KerningMode.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.Fonts
+{
+    /// <summary>
+    /// Kerning is the contextual adjustment of inter-glyph spacing.
+    /// This property controls metric kerning, kerning that utilizes adjustment data contained in the font.
+    /// </summary>
+    public enum KerningMode
```
antonfirsov, line 10: ❤️

src/SixLabors.Fonts/Tables/General/CMapTable.cs
```diff
@@ -14,29 +16,63 @@ internal sealed class CMapTable : Table
...
         {
             this.Tables = tables.OrderBy(t => GetPreferredPlatformOrder(t.Platform)).ToArray();
+            this.format14SubTables = this.Tables.OfType<Format14SubTable>().ToArray();
         }
 
         internal CMapSubTable[] Tables { get; }
 
         private static int GetPreferredPlatformOrder(PlatformIDs platform)
-            => PreferredPlatformOrder.TryGetValue(platform, out var order) ? order : int.MaxValue;
+            => PreferredPlatformOrder.TryGetValue(platform, out int order) ? order : int.MaxValue;
```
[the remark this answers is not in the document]
JimBobSquarePants: Done

src/SixLabors.Fonts/Tables/AdvancedTypographic/GPos/LigatureAttachTable.cs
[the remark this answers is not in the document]
JimBobSquarePants: Gonna leave for now. I'd like to get this in and focus on potential performance issues after.

brianpopow (approved):
I think its time to get this huge PR merged! 👍

PR #208, opened 2021-09-26
Fix loading woff 1 fonts

brianpopow:
### Prerequisites

### Description

This fixes an issue loading woff1 files: in some cases, its necessary to seek back in the stream to load strings. For that the compressed stream is read completely.


tests/SixLabors.Fonts.Tests/GlyphTests.cs
```diff
@@ -126,9 +126,32 @@ public void RenderColrGlyphTextRenderer()
...
+            Assert.True(rendererTtf.ControlPoints.SequenceEqual(rendererWoff.ControlPoints));
+        }
+
 #if NETCOREAPP3_0_OR_GREATER
         [Fact]
-        public void RenderWoff2Glyphs_EqualToTtfGlyphs()
+        public void RenderWoff2Glyphs_IsEqualToTtfGlyphs()
         {
             Font fontTtf = new FontCollection().Add(TestFonts.OpenSansFile).CreateFont(12);
             Font fontWoff2 = new FontCollection().Add(TestFonts.OpenSansFileWoff2).CreateFont(12);
```
JimBobSquarePants, line 157: I was wondering if you'd changed this as the new font will fail to load.

JimBobSquarePants (approved):
You're quick! Nice job 👍

tests/SixLabors.Fonts.Tests/GlyphTests.cs
brianpopow: I wanted to change this to the new files once we figured out the issue with the woff2 file

JimBobSquarePants: Looking at that now will push a branch shortly.

PR #210, opened 2021-09-29
Port/integrate Font hinting from SharpFont

tocsoft:
This is an initial spike at integrating the font hinting code from https://github.com/MikePopoloski/SharpFont into our solution.

This is completely untested...all the existing tests continue to pass but I've yet to even start looking at rendered output.

## TODO
- [ ] Tests
- [ ] Visually verify output
- [ ] Improve exceptions

## None goals
- optimization of hinting (there will be the option of opting out that should prevent regressions)

Targeting/based on #198 in case there was any oddities from the updated layout.


JimBobSquarePants:
Thanks for making a start on this @tocsoft 

I pushed some cleanup so the solution will build in release and fixed a couple of issues I found. I also temporarily reintroduced the constructor for `RendererOptions` so  I could see a visual output via the samples project.

Here's Tahoma at 11pt.

unihinted vs hinted.
image

Overall I can see increased crispness but there's some funkiness around `i`, `l`, `o` and `p`. 

The original doesn't seem to suffer from those issues.



The typography version seems to sit somewhere in between. 
image

One thing I've spotted that SharpFont appears to handle that ourselves and Typography do not is `CompositeGlyphFlags.RoundXYToGrid`. I wonder if that makes a difference?

src/SixLabors.Fonts/Tables/General/Glyphs/GlyphVector.cs
```diff
@@ -110,7 +129,7 @@ public static GlyphVector Append(GlyphVector first, GlyphVector second)
             }
 
             var bounds = new Bounds(first.Bounds.Min, first.Bounds.Max);
-            return new GlyphVector(controlPoints, onCurves, endPoints, bounds);
+            return new GlyphVector(controlPoints, onCurves, endPoints, bounds, first.Instructions);
```
JimBobSquarePants, line 132: I wonder if we should be appending the instructions here?

tocsoft: not that I've been able find anything to back up my theory, but my guess would be that we should be hinting each transformed composite glyph part independently and combing them post hinting/as we render.

Just munging the instruction sets together I can't see working as the instructions would have be written against the set of control points in the part not the new composite set.

JimBobSquarePants: I wonder how that would work for word though, we’d have to refactor the way we store the glyphs there. We can have a peek through the FreeType source to see if we can figure out what they do. (Hard though, the code seems scattered)

On that note I think our fallback top side bearing calculation might be off. I’ll dig out the example I found when I’m at my computer later.

src/SixLabors.Fonts/FontMetrics.cs
```diff
@@ -209,17 +209,16 @@ internal GlyphVector ApplyHinting(GlyphVector glyphVector, float pixelSize, usho
...
+            var pp2 = new Vector2(pp1.X + (advance * scale), 0);
+            var pp3 = new Vector2(0, glyphVector.Bounds.Max.Y + (verticalFrontSideBearing * scale));
+            var pp4 = new Vector2(0, pp3.Y - (verticalAdvance * scale));
 
             var controlPoints = new Vector2[glyphVector.ControlPoints.Length + 4];
-            controlPoints[controlPoints.Length - 4] = pp1 * scale;
-            controlPoints[controlPoints.Length - 3] = pp2 * scale;
-            controlPoints[controlPoints.Length - 2] = pp3 * scale;
-            controlPoints[controlPoints.Length - 1] = pp4 * scale;
+            controlPoints[controlPoints.Length - 4] = pp1;
```
tocsoft, line 218: nice... this would explain the odd `p`s and `i` etc

src/SixLabors.Fonts/Tables/General/Glyphs/GlyphVector.cs
JimBobSquarePants: image


@tocsoft Looks like you're on to something.

JimBobSquarePants:
Fixed up the bearing calculations to match FreeType. Even better now to my eye.

image
image

JimBobSquarePants:
@tocsoft  Working on a plan for composite glyphs. Watch this space.

JimBobSquarePants:
I've gotten composite working by doing what FreeType do and ignore composite instructions, hinting each entry separately. 

Here's an example of the ligature `fi` - `\uFB01` rendered with Tahoma at 11.25pt.

Hinted on the left.
image

image

However, I'm not 100% happy with the performance (_I made the entries immutable when I should have bothered since we use `DeepClone` where we need to_) and I also have 5 tests comparing composite glyphs to Woff2 failing (_I need to investigate that._)

Rather than pushing directly to this PR I created a new branch [js/font-hinting-composite](https://github.com/SixLabors/Fonts/tree/7abda0eb063a9db9af423de16e68f25d01209ba3) so you could have a look at the changes there first

brianpopow:
> I also have 5 tests comparing composite glyphs to Woff2 failing (_I need to investigate that._)
> 

For Font hinting we need the glyph instructions, right? That is still a TODO in loading the Woff2 font format.

JimBobSquarePants:
>That is still a TODO in loading the Woff2 font format.

That's probably that then! 😂

JimBobSquarePants:
@tocsoft I ended up merging my code once I'd fixed the performance issues. Composite glyphs are now hinted by hinting each sub glyph.

Output looks great to me.

Open Sans 11.25pt/15px
ﬁ ttf
ﬁ woff

Tahoma 11.25pt/15px
Lorem ipsum dolor sit amet

tocsoft:
Thats fine... during my testing over the weekend on this branch i've seen a couple issues.

Looking closely at the hinted text, `i`s and `l` seem to get moved over by a pixel, which isn't very noticeable unless you look at it at about the 8pt mark using Tahoma, in which case it goes from readable to overlapping other letters.

we have separate issue with the text renderer in IS.Drawing when re rasterize with anti-aliasing turned off (again @ 8pt Tahoma) we still get some anti-aliased pixels anyway (this is just a bug in out vector rasterization where we end up over writing the dirty state of scanline thus not normalizing for anti-aliased text... I'll try doing a spike updating the anti-aliased code path of the rasterizer to sort this... might even try and in an option to handle pixel centre based rendering in place of percentage covered based raster and see if that produces any better results (obviously mixed with hinting).

JimBobSquarePants:
>Looking closely at the hinted text, is and l seem to get moved over by a couple of pixels, which isn't very noticeable unless you look at it at about the 8pt mark using Tahoma, in which case it goes from readable to overlapping other letters.

I can have a look at the hinter itself and compare the instruction handling with FreeType

JimBobSquarePants:
@tocsoft So FreeType has a whole bunch of custom rules for individual fonts where it does adjustments prior to hinting.

These rules define horizontal scaling tweaks (_1000 == 1.0_)

image

Oddly though it seems that the rule is specific to particular point sizes not an upper limit. (_This makes no sense to me as error will increase the smaller the size_)

image

JimBobSquarePants:
@tocsoft I'm actually gonna go ahead and merge this as it's gonna block the options fixes and will be difficult to maintain in a separate branch. 

I've made the default `ApplyHinting` property false to avoid surprises. I'll continue to look at it once I'm done elsewhere.

JimBobSquarePants (approved):

PR #218, opened 2021-11-12
Introduce TextOptions

### LINKED ISSUE #212: Issues with RendererOptions
- The type name is confusing in ImageSharp.Drawing context. There are many many `*Options` types across our libraries, most of them indicating what is the type about. Here it's unclear at the first glance what are we rendering. 
- Some properties are get only and initialized from the constructor while others are settable.
- We have a `TextOptions` type there that duplicates a subset of the functionality which is confusing. Upstream issue: https://github.com/SixLabors/ImageSharp.Drawing/issues/174

JimBobSquarePants:
### Prerequisites

### Description

Introduces `TextOptions` which replaces `RendererOptions` in this library and `TextOptions` in ImageSharp.Drawing to create a seamless and simplified API for consumers. Fixes #212 

The new type simplifies the constructors and introduces the validation from the downstream equivalent.

Downstream the `TextOptions` type will be removed and will no longer be referenced as a property by the global `DrawingOptions`. See this [comment](https://github.com/SixLabors/Fonts/issues/212#issuecomment-965923198) for reasoning. As a result syntactic changes for altering default property values in the options will be cleaner. 

e.g. Oddities like this will be gone where we are creating nested types and passing what should be properties of a single type in two separate ways. 

Old
``` c#
var options = new DrawingOptions {
    TextOptions = new TextOptions { ApplyKerning = false }
};

... DrawText(options, text, font, brush, origin);
```

New
``` c#
TextOptions options = new(font) {
    KerningMode = KerningMode.None,
    Origin = origin,
};

... DrawText(options, text, brush);
```

This removes the odd internal copying behavior and duplication of properties (_that, incidentally, makes development of features touching the types incredibly difficult_). Consumers now have full access to the layout properties from a single source. The `Font` parameter is required by the constructor as that is the only property without a default value as as such is required. 

I've been running a local build of ImageSharp.Drawing using the new approach and it is much cleaner. I'll be creating a PR once this is merged.

_Note: Ignore the Drawing  samples code in this PR. They're hacked so that the code compiles. I'll do a follow up PR to cleanup once ImageSharp.Drawing is updated._


src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -543,22 +543,8 @@ private static void SubstituteBidiMirrors(FontMetrics fontMetrics, GlyphSubstitu
                         continue;
                     }
 
-                    CodePoint codePoint = codePointEnumerator.Current;
```
JimBobSquarePants, line 546: This fixes an issue discovered downstream where we were incorrectly trimming whitespace from the beginning of new lines following a mandatory line break.

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -665,7 +653,10 @@ private static void SubstituteBidiMirrors(FontMetrics fontMetrics, GlyphSubstitu
                     }
 
                     // Do not start a line following a break with breaking whitespace
```
JimBobSquarePants, line 655: This fixes an issue discovered downstream where we were incorrectly trimming whitespace from the beginning of new lines following a mandatory line break.

antonfirsov:
My main concern from #212 discussions was the potential duplication of `Font`. Based on the sample code it looks like you are planning to mitigate this by removing the `Font` parameter from overloads asking for `DrawingOptions`, so there's no chance do calls as below. Do I get it right?

```C#
TextOptions options = new(FONT_1) {
    KerningMode = KerningMode.None,
    Origin = origin,
};

... DrawText(options, FONT_2, text, brush);

```

> I've been running a local build of ImageSharp.Drawing using the new approach and it is much cleaner. I'll be creating a PR once this is merged.

Any chance you can create a (non compiling) WIP PR, so we can see the changes in ImageSharp.Drawing? This is where the real impact is, it's hard to reason about this change looking at the Fonts library only.

JimBobSquarePants:
>Do I get it right? 
@antonfirsov  Yep. No possibility of that at all. I've refactored the extensions to ensure that. 

>Any chance you can create a (non compiling) WIP PR, so we can see the changes in ImageSharp.Drawing? This is where the real impact is, it's hard to reason about this change looking at the Fonts library only.

Sure, yeah, here's a branch I have using the new API. I'm referencing the Fonts library as a project to allow local compilation for me.

https://github.com/SixLabors/ImageSharp.Drawing/compare/SixLabors:e987bd8...SixLabors:a3ecc76

Here's an example of nested options which I want to remove. https://github.com/SixLabors/ImageSharp.Drawing/compare/SixLabors:e987bd8...SixLabors:a3ecc76#diff-603320c5c2cf99caf8b8dca6b0923f7542ead47c3540656532925da6d615fa57L49-L59

This copying here from the original subset was killing us during font feature development and needs to be removed. I had to keep setting non-default values for properties just to see some visual output. Plus of course the feature coupling is circular, you have to add properties in two locations to use anything new.
https://github.com/SixLabors/ImageSharp.Drawing/compare/SixLabors:e987bd8...SixLabors:a3ecc76#diff-20db11dc06f25f4deceb1052b40e4e05ac93f38868c8cd4333c43eb833737b71L47-L58

JimBobSquarePants:
@SixLabors/core what do you think of this?

src/SixLabors.Fonts/TextOptions.cs
```diff
@@ -0,0 +1,172 @@
...
+            this.VerticalAlignment = options.VerticalAlignment;
+            this.LayoutMode = options.LayoutMode;
+            this.KerningMode = options.KerningMode;
+            this.ColorFontSupport = options.ColorFontSupport;
+        }
+
+        /// <summary>
+        /// Gets or sets the font.
+        /// </summary>
+        public Font Font { get; set; }
```
antonfirsov, line 56: A project consuming Fonts without `<Nullable>enable</Nullable>` can actually set this to null, so adding guards is still necessary if we assume in Fonts and Drawing code that it's never null.

Same concern around constructor.

src/SixLabors.Fonts/TextOptions.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        /// <summary>
+        /// Gets or sets the collection of fallback font families to use when
+        /// a specific glyph is missing from <see cref="Font"/>.
+        /// </summary>
+        public ICollection<FontFamily> FallbackFontFamilies { get; set; } = new HashSet<FontFamily>();
```
antonfirsov, line 62: I would avoid backing this with a heavy type like `HashSet<FontFamily>()`. How do we expect this to be consumed?
Can't it be just `IReadOnlyList<FontFamily>`, so it's `Array.Empty<FontFamily>` by default, and users can set it like:

```C#
textOptions.FallbackFontFamilies  = new FontFamily[] { ff1, ff2, ff3 };
```

antonfirsov (commented):
No major concerns around the API proposal and Drawing impact at first glance. We can come back and iterate further if we find any issues.

Can't say anything about implementation details though. Hoping that @tocsoft and @brianpopow can also have a look.

src/SixLabors.Fonts/TextOptions.cs
JimBobSquarePants: Yeah so I wanted to use `ICollection<FontFamily>` backed by `HashSet<FontFamily>` for two reasons.

1. I didn't want duplicate `FontFamily` in the collection. It's not harmful just wasteful. (I can probably deal with that in `TextLayout` though)
2. I wanted something we can immediately add to and continue to do so. However upon reflection that can be confusing since we can set the property itself. 

I'll update. You approach is better.

src/SixLabors.Fonts/TextOptions.cs
JimBobSquarePants: Good point, Will update.

src/SixLabors.Fonts/TextOptions.cs
JimBobSquarePants: Done

src/SixLabors.Fonts/TextOptions.cs
JimBobSquarePants: Done

PR #223, opened 2021-11-29
Allow returning names by id.

### LINKED ISSUE #222: Change NameId to be public and add method in FontDescription to be able to access the private NameTable
### Discussed in https://github.com/SixLabors/Fonts/discussions/206

<div type='discussions-op-text'>

<sup>Originally posted by **beeradmoore** September 11, 2021</sup>
Recently I was having an issue of getting the font name in the format I expecting to find. I found that the type of name is called a postscript name. Looking through the source I found the enum which would allow me to access it in [NameIds](https://github.com/SixLabors/Fonts/blob/87b91260a2e9fdde948f9b447856315f43f13eb1/src/SixLabors.Fonts/WellKnownIds/NameIds.cs#L47).

Looking into how `FontDescription` interacts with a `NameId` (for an existing call that does work) I can see that it calls [FontName](https://github.com/SixLabors/Fonts/blob/87b91260a2e9fdde948f9b447856315f43f13eb1/src/SixLabors.Fonts/FontDescription.cs#L66) which then calls [FontName ](https://github.com/SixLabors/Fonts/blob/97e3ed651f0c8f365ef7aebae88a168f33b94ae6/src/SixLabors.Fonts/Tables/General/NameTable.cs#L38) on the private `NameTable` which would then call [GetNameById](https://github.com/SixLabors/Fonts/blob/97e3ed651f0c8f365ef7aebae88a168f33b94ae6/src/SixLabors.Fonts/Tables/General/NameTable.cs#L59).

For my quick fix to get an app out for testing on a Friday afternoon I just added my own `PostscriptName` method which follows the same pattern as above.

I propose that `NameId` enum should not be internal, and `FontDescription` should have a `GetNameById` so that the user can get any property that is already loaded into the `NameTable`.

This was first mentioned in discussion #155.</div>

JimBobSquarePants:
### Prerequisites

### Description

Adds a new `GetNameById(CultureInfo culture, KnownNameId nameId)` to `FontDescription` and exposes the `KnownNameId`  enum. Fixes #222 

CC @beeradmoore


tocsoft (approved):
lgtm 👍

PR #225, opened 2021-12-04
Optimize hinting interpreter

TechnologicalPizza:
### Prerequisites

### Description

Optimized the hinting interpreter to generate better JIT assembly by 
  * creating throw helpers.
  * reducing calls to `ReadOnlyMemory<byte>.Span` and eliding bound checks when reading bytes.
  * not reading bytes when skipping.


TechnologicalPizza:
Did a crude benchmark by running drawing the same text 1000 times.

Before:
before

After:
after

JimBobSquarePants (approved):
LGTM. Thanks!👍

We actually need to do a proper cleanup of this type at some point and also investigate customized hinting adjustment for individual fonts.

https://github.com/SixLabors/Fonts/pull/210#issuecomment-934333756

PR #230, opened 2022-01-23
Fix rendering for fully-qualified  emojis with Twemoji font

jairbubbles:
### Prerequisites

### Description

Some emojis can be fully-qualified with a variation selector (`u+FE0F`) to differentiate them from their text counterpart.

Ex (from https://www.unicode.org/Public/emoji/14.0/emoji-test.txt):
```
263A FE0F                                              ; fully-qualified     # ☺️ E0.6 smiling face
```

With Twemoji Mozilla font the variation selector was rendered:

I modified the code to completely ignore all variation selectors (no only the emoji one) when layouting:


src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -554,11 +554,18 @@ private static void SubstituteBidiMirrors(FontMetrics fontMetrics, GlyphSubstitu
...
+                        codePointIndex++;
+                        graphemeCodePointIndex++;
+                        continue;
+                    }
+
                     if (CodePoint.IsTabulation(codePoint))
                     {
                         glyphAdvance *= options.TabWidth;
                     }
-                    else if (CodePoint.IsZeroWidthJoiner(codePoint) || CodePoint.IsZeroWidthJoiner(codePoint))
```
JimBobSquarePants, line 561: I think you found a type there. The second condition should be `CodePoint.IsZeroWidthNonJoiner(codePoint)` as I recall.

JimBobSquarePants:
Thanks for this! 

I think you've definitely found the correct part in the pipeline to do the filtering as I would assume substitution would deal with handling of text variations if possible. 

BTW, best explanation of sequences I found was here. https://www.emojiall.com/en/help-emoji-variation-sequences

If we can fix up that condition then I'll get this merged.

src/SixLabors.Fonts/TextLayout.cs
jairbubbles: Updated! 

(I'm wondering it we should completely avoid them and not generate glyphs for them too)

JimBobSquarePants: As I recall they affect bidi reordering so must be kept.

JimBobSquarePants (approved):

jairbubbles:
Thx @JimBobSquarePants!

FYI I'm working on adding support for Emojis in a screenshot utility => see https://github.com/greenshot/greenshot/pull/358

PR #231, opened 2022-01-30
Fix some emoji rendering with Segoe UI Emoji font

jairbubbles:
### Prerequisites

### Description

When rendering some emojis the text was not measured properly.

For instance "women holdings hands" was ok but not some of its variants:
image

```csharp
var fonts = new FontCollection();
FontFamily segoeEmoji = fonts.Add(@"C:\Windows\Fonts\seguiemj.ttf");

// 1F469 1F3FB 200D 1F91D 200D 1F469 1F3FC; # 👩🏻‍🤝‍👩🏼 E12.1 women holding hands: light skin tone, medium-light skin tone
RenderText(segoeEmoji, "\U0001F469\U0001F3FB\u200D\U0001F91D\u200D\U0001F469\U0001F3FC", pointSize: 72); // => not good (right part of the emoji is ignored)

// 1F46D 1F3FB   # 👭🏻 E12.0 women holding hands: light skin tone
RenderText(segoeEmoji, "\U0001F46D\U0001F3FB", pointSize: 72); // => good
```
After investigation it appears that the substitution is generating glyphs on the code point `\u200D` (Zero width Joiner) and its width is ignored during layout.

My fix tend to detect such a case and not ignore ZWJ when it's the result of a substitution.

image

*I'm not sure of the fix and I need guidance on how to write tests as I need `Segoe UI Emoji` which I believe can't be submitted on this repository so I'm keeping this as draft for the moment*


src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -565,11 +565,11 @@ private static void SubstituteBidiMirrors(FontMetrics fontMetrics, GlyphSubstitu
...
                         glyphAdvance *= options.TabWidth;
                     }
-                    else if (CodePoint.IsZeroWidthJoiner(codePoint) || CodePoint.IsZeroWidthNonJoiner(codePoint))
+                    else if (metrics.Length == 1 && (CodePoint.IsZeroWidthJoiner(codePoint) || CodePoint.IsZeroWidthNonJoiner(codePoint)))
                     {
                         // The zero-width joiner characters should be ignored when determining word or
-                        // line break boundaries so are safe to skip here.
-                        // Any existing instances are the result of font error.
+                        // line break boundaries so are safe to skip here. Any existing instances are the result of font error.
+                        // It multiple metrics are associated with code point, they are most likely the result of a substitution so we shouldn't ignore it.
```
JimBobSquarePants, line 572: Gotta love when name and behavior doesn't match eh? 

This fix looks good to me. As far as unit testing goes, assuming we cannot use the Mozilla Font, we'll have to create a Windows only test and reference the font from the system collection. I'm not sure we can legally download and include it. [Distribution](https://docs.microsoft.com/en-us/typography/font-list/segoe-ui-emoji#products-that-supply-this-font) covered 10 and 11 so the CI images should be good.

You can wrap the test in the following compiler conditional `#if OS_WINDOWS`

jairbubbles: It's not exactly the same use but in the FAQ it says:
>  Web fonts are fonts that are hosted on a web server. You do not have rights to:
>  -  copy fonts from a Windows installation to a web server, a process known as web font “self-hosting”.
>  - convert the font to the formats typically associated with web fonts, such as the WOFF or WOFF2 format.

> You can wrap the test in the following compiler conditional #if OS_WINDOWS

The problem is that the font is very different between Windows 10 and Windows 11. I just tested and some emojis are broken with the old font. PR incoming 😅

JimBobSquarePants: Thanks for your help! Maybe we can get away with it then, the language appears to allow our use case. If anyone complains we can take it down

jairbubbles: > If anyone complains we can take it down

I like your pragmatism! 😉

jairbubbles: I added a simple unit test. I noticed that the width is not exactly the same. I'm not sure if it's the font or if there's still an issue on how the width is computed.

JimBobSquarePants (approved):

PR #234, opened 2022-02-09
Drop NetStandard 1.3

JimBobSquarePants:
### Prerequisites

### Description

From the [Microsoft Documentation](https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0#which-net-standard-version-to-target)

>We recommend you target .NET Standard 2.0, unless you need to support an earlier version. Most general-purpose libraries should not need APIs outside of .NET Standard 2.0. .NET Standard 2.0 is supported by all modern platforms and is the recommended way to support multiple platforms with one target.

There's no need for us to support this.  .NET Standard 2.0 is supported by pretty much everything.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -20,36 +20,20 @@
 
   <!--
     https://apisof.net/
-    +===================+=======+==========+=====================+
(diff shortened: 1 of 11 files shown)


brianpopow (approved):
👍

PR #241, opened 2022-02-20
Make HasPathTests a theory instead of a fact

0xced:
Having the whole list of system fonts gives a better view of what is actually being tested.

Also don't recreate a new SystemFontCollection instance for each test in the FontFamilyTests class but use the lazily evaluated SystemFonts.Families properties instead.


JimBobSquarePants (approved):
Yeah, much better. Thanks! 👍

PR #251, opened 2022-03-02
Support text runs.

### LINKED ISSUE #10: rendering engine should be able to support spans of characters with different formats
Provide a mechanisum for marking up spans of the supplied text with different formats allowing for simple rich text support

i.e. The **quick** brown fox jumps _over_ the lazy dog
would use the below proposed font style
```c#
new SpanningFontStyle(
    new Span(5, 9){ bold = true; },
    new Span(28,31){ italic = true; },
);
```

JimBobSquarePants:
### Prerequisites

### Description

Adds initial support for individual text runs within a text shaping run. Fixes #10

TODO: 
- [x] Fix Font size application to individual metrics. 
- [x] Capture subscript/superscript/strikethrough metrics for attributes from [OS2 table](https://docs.microsoft.com/en-us/typography/opentype/spec/os2) so we can render.
- [x] Capture underline metrics for attributes from [Post table](https://docs.microsoft.com/en-us/typography/opentype/spec/post) so we can render.
- [x] Implement rendering in `GlyphMetrics`

Example API for setting `TextRun` properties. `Start` and `End` properties are measured in graphemes to avoid slicing up emoji and such like. 

```c#
var textRuns = new List<TextRun>
{
    new TextRun { Start = 4, End = 9, Font = uiFont.CreateFont(32, FontStyle.Bold), TextAttributes = TextAttribute.Underline | TextAttribute.Strikethrough },
    new TextRun { Start = 26, End = 30, Font = uiFont.CreateFont(32, FontStyle.Italic) }
};
```

Example Output.

The quick brown fox jumps over the lazy dog


src/SixLabors.Fonts/TextOptions.cs
```diff
@@ -9,9 +9,9 @@
 namespace SixLabors.Fonts
 {
     /// <summary>
-    /// Provides configuration options for rendering text.
+    /// Provides configuration options for rendering and shaping text.
     /// </summary>
-    public sealed class TextOptions
+    public class TextOptions
```
JimBobSquarePants, line 14: I've kept `TextOptions` as the name here. I want to use `TextDrawingOptions` in ImageSharp.Drawing

src/SixLabors.Fonts/TextOptions.cs
```diff
@@ -181,5 +182,10 @@ public float LineSpacing
         /// Gets or sets the collection of additional feature tags to apply during glyph shaping.
         /// </summary>
         public IReadOnlyList<Tag> FeatureTags { get; set; } = Array.Empty<Tag>();
+
+        /// <summary>
+        /// Gets or sets an optional collection of text runs to apply to the body of text.
+        /// </summary>
+        public IReadOnlyList<TextRun> TextRuns { get; set; } = Array.Empty<TextRun>();
```
JimBobSquarePants, line 189: `IReadOnlyList<T>` still allows covariance and is more efficient.

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -388,18 +385,78 @@ private static IReadOnlyList<GlyphLayout> LayoutText(TextBox textBox, TextOption
...
+                        Font = options.Font
+                    }
+                };
+            }
+
+            int start = 0;
+            int end = CodePoint.GetCodePointCount(text);
+            List<TextRun> textRuns = new();
+            foreach (TextRun textRun in options.TextRuns!.OrderBy(x => x.Start))
+            {
```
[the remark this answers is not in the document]
JimBobSquarePants: Yep, that was coming. Following text runs will now update the end point of previous runs if overlapping.

JimBobSquarePants:
@tocsoft My reasoning re `TextOptions` was that the new `TextRun` override in Drawing would be `TextDrawingRun`

so the following naming looked correct.
- `TextOptions` - `TextRun`
- `TextDrawingOptions` - `TextDrawingRun`

It also made development a little chicken and eggy since the referenced Drawing version in the Samples project would not build if I change the name.

JimBobSquarePants:
Font size now works for each run.

I am big and I am small

JimBobSquarePants:
@tocsoft I've captured metrics for subscript/superscript and strikethrough that can be used for rendering. My assumption would be that this is performed in `GlyphMetrics.RenderTo()`. 

I'm gonna need some help with the actual render implementation though, drawing is still not a strong point of mine.

Still need to capture underline metrics.

tocsoft:
> @tocsoft I've captured metrics for subscript, superscript, underline, and strikethrough that can be used for rendering. My assumption would be that this is performed in `GlyphMetrics.RenderTo()`.
> 
> I'm gonna need some help with the actual render implementation though, drawing is still not a strong point of mine.

Yeah the rendering side I should be able to tackle... ultimately for all this we will need to pass long a `TextRun` (either synthetic or user created) long to the `IGlyphRenderer` specifically we should be able to add a new `TestRun` field to `GlyphRendererParameters` with the applicable `TextRun` for the glyph, we can then utilise that on the Drawing side to handle things like changes to the Brush/Pen, and knowing to draw the strike throughs/underlines etc.

`GlyphRendererParameters` is also where we can access the metrics for underlining etc.

JimBobSquarePants:
@tocsoft I'm _pinches fingers_ "this close" to cracking superscript and subscript rendering. I've got scaling and positioning working, I just need to fix an indexing issue with the runs (_exclusive end turned out to be a bad idea._) so hold out please.

JimBobSquarePants:
I'll push the updates soon

superscript and subscript

JimBobSquarePants:
Found some issues with Arabic scripts where it's common to find glyphs with zero advance (Width/Height). Working on a fix.

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -14,48 +14,99 @@ namespace SixLabors.Fonts
...
             }
 
-            if (options.WrappingLength > 0)
+            TextBox textBox = ProcessText(text, options);
+            return LayoutText(textBox, options);
+        }
+
+        public static IReadOnlyList<TextRun> BuildTextRuns(ReadOnlySpan<char> text, TextOptions options)
+        {
+            if (options.TextRuns?.Count == 0)
```
tocsoft, line 32: This doesn't work as expected if `options.TextRuns` is null. if `options.TextRuns` is null here then it still doesn't equal `0` as it equals `null` instead.

src/SixLabors.Fonts/GlyphMetrics.cs
```diff
@@ -194,107 +234,175 @@ internal FontRectangle GetBoundingBox(Vector2 origin, float scaledPointSize)
...
+                    float top = position - (height * .5F);
+                    float bottom = top + height;
+
+                    Vector2 scale = new Vector2(scaledPoint) / this.ScaleFactor * MirrorScale;
+                    Vector2 offset = location + (this.offset * scale * MirrorScale);
+
+                    // Calculate the correct advance for the line.
+                    // For zero advance glyphs we must calculate our own advance.
+                    // We add a single font unit to account for rounding error.
+                    float lsb = Math.Abs(this.LeftSideBearing) + 1;
```
JimBobSquarePants, line 360: @tocsoft I had to do a bit of tweaking of your code (and following fix) to get this example working. There were still gaps in the output line strokes. I'm curious to see if it looks ok with different colored brushes.


بِسْمِ ٱللَّهِ ٱلرَّحْمَٟنِ ٱلرَّحِيمِ
بِسْمِ ٱللَّهِ ٱلرَّحْمَٟنِ ٱلرَّحِيمِ


Turns out we had a pretty major error in the `TextMeasurer` also where we weren't taking into consideration vertical offset of marks. The Segoe UI version was horribly cut off.

src/SixLabors.Fonts/GlyphPositioningCollection.cs
```diff
@@ -216,19 +237,28 @@ private bool Add(FontMetrics fontMetrics, GlyphSubstitutionCollection collection
...
             ushort glyphId = data.GlyphIds[0];
-            foreach (GlyphMetrics m in this.map[this.offsets[index]])
+            foreach (GlyphMetrics m in this.map[this.offsets[index]].Metrics)
             {
                 if (m.GlyphId == glyphId && fontMetrics == m.FontMetrics)
                 {
-                    m.ApplyOffset((short)data.Bounds.X, (short)data.Bounds.Y);
-                    m.SetAdvanceWidth((ushort)data.Bounds.Width);
-                    m.SetAdvanceHeight((ushort)data.Bounds.Height);
+                    if (isDirtyXY)
```
JimBobSquarePants, line 252: I split out dirty checks here because both values can be adjusted separately.

src/SixLabors.Fonts/GlyphMetrics.cs
```diff
@@ -194,107 +234,175 @@ internal FontRectangle GetBoundingBox(Vector2 origin, float scaledPointSize)
         /// <param name="location">The location.</param>
         /// <param name="options">The options used to influence the rendering of this glyph.</param>
         /// <exception cref="NotSupportedException">Too many control points</exception>
-        public void RenderTo(IGlyphRenderer surface, float pointSize, Vector2 location, TextOptions options)
+        internal void RenderTo(IGlyphRenderer surface, float pointSize, Vector2 location, TextOptions options)
```
JimBobSquarePants, line 237: Making this internal stops users from attempting to render glyphs without an autogenerated run. I've streamlined the process to use `Font.GetGlyphs(CodePoint codePoint, ColorFontSupport support)` which pushes users through the appropriate pipeline.

src/SixLabors.Fonts/GlyphMetrics.cs
JimBobSquarePants: Line height is correct in both. The first font is a specialist one designed specifically for reading the Quran.

JimBobSquarePants:
@tocsoft I tested the codase at https://github.com/SixLabors/Fonts/pull/251/commits/d30b285fd5fe2c15d69a41183534f88b4e333490 with your ImageSharp.Drawing PR and the strokes are now pixel perfect starting and stopping at the exact glyph position. 

DrawRichTextWithMixOfPensAndBrushes_Solid500x200_(0,0,0,255)_RichText-F(32)
DrawRichTextWithMixOfPensAndBrushes2_Solid500x200_(0,0,0,255)_RichText-F(32)

tocsoft:
awesome... I did notice the line hights end up funky on the 2nd line when you have inline font size changes.

I would expect that 2nd line much closer then that as it doesn't have the taller text.

JimBobSquarePants:
>I would expect that 2nd line much closer then that as it doesn't have the taller text. 

The line height is governed by the maximum line height of all glyphs (to avoid line collision) so the first line will have a much larger line height therefore pushing the second one down.

Here's what MS word does at 32pt using Calibri, It's virtually identical.

image

I wanted to ask actually, What do you think of the possibility of adding an extra step in-between `IGlyphRenderer.BeginGlyph` and `IGlyphRenderer.EndGlyph` and move the underline figure to there? I'm thinking that could allow us to render underlines with different colors like MS Word can.

tocsoft:
> I wanted to ask actually, What do you think of the possibility of adding an extra step in-between `IGlyphRenderer.BeginGlyph` and `IGlyphRenderer.EndGlyph` and move the underline figure to there? I'm thinking that could allow us to render underlines with different colors like MS Word can.

I did have a similar thought... I actually think we might want a completely different (additional) API that allows passing along the start/end line + thickness rather than drawing the rectangle... That would allow better pen support for strikeouts. I'm thinking it should be an optional interface like with `IColorGlyphRenderer` something like `ITextDecorationRenderer` (for example) would inherit `IGlyphRenderer` and it would expose `DrawDecoration(Vector2 start, Vector2 end, float thickness, TextDecorationType TextRun textRun)`

with `TextDecorationType` being an enum exposing `Strikeout` and `Underline` (would allow for expanding to `Overline` too in the future too).

The benefit of this API is that it would allow us to merge decorations across glyphs/runs and allow completely removing an issues with gaps between them and also full `Pen` support formats and all (I would need to think of best way to support dynamic pen thickness, maybe add support for a 'default'/unset pen thickness convention)

The idea would be that we would use the `DrawDecoration` API if its available and fallback to our current implementation if not. 

I'll have to have a play with it over in the drawing branch and see if I can't land on something that feels good.

tocsoft:
I've split out `TextAttributes` and `TextDecorations` into 2 enums as they target different parts of the stack, one effects the layout side and the other is just additional decorations on top.

But we can now override the pen used while drawing decorations.

DrawRichTextRainbow_Solid500x300_(0,0,0,255)_RichText-ArabicRainbow-F(40)
DrawRichText_Solid500x300_(0,0,0,255)_RichText-F(40)

JimBobSquarePants:
@tocsoft  Just did a BIG update to the layout engine to ensure we generate the correct line-height + offsets for mixed font sizes.

Here's how we compare to Chromium based browsers rendering mixed font sizes. It's now exact.

Horizontal

layout-horizontal

Vertical

layout-vertical

Found and fixed a line-break issue in the process and also vertical and horizontal centering was a shade off. Test updates have also been verified visually.

JimBobSquarePants:
Too tired to fix the .NET 472 rounding silliness just now. Will fix tomorrow.

JimBobSquarePants:
@tocsoft I'm pretty happy with the shape of this now and with your blessing will get it merged. The last things I wanted to have a look at was font hinting and It looks like I've solved the horizontal alignment problem of the `l` and `i` glyphs

Tahoma 11.25pt @ 72 dpi

Unhinted
Unhinted

Hinted
Hinted

Tahoma 11.25pt @ 96 dpi

Unhinted
Unhinted

Hinted
Hinted

tocsoft (approved):

PR #278, opened 2022-06-19
Make TableLoader works without reflection

kant2002:
I did try to port https://github.com/JamesRandall/fsharp-wolfenstein to NativeAOT in reflection-free mode, to see how far I can go, and this is only piece which prevents me to use NativeAOT without reflection in this library.

I notice comment in the Register abotu static interface methods, but seems to be manual registration can be used fine, and seems to be less magic.


src/SixLabors.Fonts/Tables/TableLoader.cs
```diff
@@ -27,29 +27,29 @@ internal class TableLoader
...
+            this.Register(VerticalMetricsTable.TableName, VerticalMetricsTable.Load);
+            this.Register(MaximumProfileTable.TableName, MaximumProfileTable.Load);
+            this.Register(OS2Table.TableName, OS2Table.Load);
+            this.Register(IndexLocationTable.TableName, IndexLocationTable.Load);
+            this.Register(GlyphTable.TableName, GlyphTable.Load);
+            this.Register(KerningTable.TableName, KerningTable.Load);
+            this.Register(ColrTable.TableName, ColrTable.Load);
+            this.Register(CpalTable.TableName, CpalTable.Load);
+            this.Register(GPosTable.TableName, GPosTable.Load);
+            this.Register(GPosTable.TableName, GPosTable.Load);
```
dlemstra, line 45: This should be GSubTable?

kant2002: Thank you for catching this! Too much fast and manual replacement

dlemstra (commented):
Looks like a good improvement @JimBobSquarePants to remove this reflection. I think we could also remove the `TableNameAttribute` attribute or do you have future plans with this?

JimBobSquarePants:
No plans, attribute can go. This is a great improvement 👍

JimBobSquarePants (approved):
Thanks for this! 👍

PR #280, opened 2022-06-22
Prevent crash from Billion Laughs attack.

JimBobSquarePants:
### Prerequisites

### Description

See

https://rawgit.com/unicode-org/text-rendering-tests/master/reports/fontkit.html#GSUB-3
https://en.wikipedia.org/wiki/Billion_laughs_attack
https://github.com/foliojs/fontkit/issues/221


tests/SixLabors.Fonts.Tests/Tables/AdvancedTypographic/Gsub/GSubTableTests.cs
```diff
@@ -368,5 +368,15 @@ public void OldStyleFiguresFeature_Works()
...
         }
+
+        [Fact]
+        public void BillionLaughsAttackDoesNotThrowException()
+        {
+            // Arrange
+            Font font = new FontCollection().Add(TestFonts.GSubLookupType2BillionLaughs).CreateFont(12);
+
+            // Act
+            TextMeasurer.Measure("lol", new TextOptions(font));
```
brianpopow, line 379: While text measure will not throw hear (or hang indefinitely), it still gives a ridiculous large number for width, which can cause an exception from the memory allocator if the user try's to render this. Should we try to address this also or should the user be responsible for checking sane values of width and height before rendering?

JimBobSquarePants: Looks like I’m using the wrong constant. That one is for subsetting. Will update later.

JimBobSquarePants: Done. We check both the length and the number of operations for GSUB, GPOS, & KERN.

brianpopow: That's much better, also a good idea to use also in GPOS and KERN.

brianpopow (approved):

Issue #268, opened 2022-05-28
Measuring text with `TextMeasurer.Measure` is slow


0xced:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of Fonts
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/Fonts/issues) and [closed](https://github.com/SixLabors/Fonts/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

I tried to use SixLabors.Fonts to [compute widths](https://github.com/0xced/Simplexcel/blob/AdjustWidthToContents/src/Simplexcel.TestApp/WorksheetExtensions.cs) of many different texts to produce a nice looking Excel file with [Simplexcel](https://github.com/mstum/Simplexcel). It turned out to be an unusable solution because it was too slow. I tracked the slowness down to the `SixLabors.Fonts.TextMeasurer.Measure` function.

I ran some benchmarks, comparing `SixLabors.Fonts` to `SkiaSharp` and here are the results.

```ini
BenchmarkDotNet=v0.13.1, OS=macOS Catalina 10.15.7 (19H1615) [Darwin 19.6.0]
Intel Core i9-9980HK CPU 2.40GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK=6.0.300
  [Host]    : .NET 6.0.5 (6.0.522.21309), X64 RyuJIT
  MediumRun : .NET 6.0.5 (6.0.522.21309), X64 RyuJIT

Job=MediumRun  IterationCount=15  LaunchCount=2  
WarmupCount=10  
```

|         Method |                 Text |         Mean |       Error |      StdDev |
|--------------- |--------------------- |-------------:|------------:|------------:|
| **SixLaborsFonts** |          **Hello world** |  **31,270.4 ns** |   **621.99 ns** |   **871.94 ns** |
|      SkiaSharp |          Hello world |     539.0 ns |     8.13 ns |    11.92 ns |
| **SixLaborsFonts** | **The q(...)y dog [43]** | **110,846.6 ns** | **4,511.26 ns** | **6,752.24 ns** |
|      SkiaSharp | The q(...)y dog [43] |   1,421.4 ns |    38.72 ns |    55.54 ns |
| **SixLaborsFonts** |                    **a** |   **7,025.6 ns** |   **349.17 ns** |   **511.81 ns** |
[...]

PR #281, opened 2022-06-24
Speed up Unicode Data lookups and remove reflection.

### LINKED ISSUE #279: UnicodeData class is not working without reflection
Simple call to `UnicodeData.GetBidiData` produce error in NativeAOT reflection-free mode. Even reflection-free mode is niche, would be good somehow provide workaround for these scenarios.

Underlying issues is 
https://github.com/SixLabors/Fonts/blob/034a440aece357341fcc6b02db58ffbe153e54ef/src/SixLabors.Fonts/Unicode/UnicodeData.cs#L45-L48

Embedded resources unlikely will even work in reflection-free mode for NativeAOT so maybe would be possible provide some additional API which provide additional sources for these data, maybe from extenally supplied `Stream`?

JimBobSquarePants:
### Prerequisites

### Description

This PR does a couple of things.

- Removes reflection from UnicodeTrie reading. Fixes #279 
- Removes bounds checks when performing Unicode property lookups. Touches #268 
- Adds a benchmark project

I chose to remove the compression from the trie packing for performance since the binaries do not grow out of hand and the benefit is great. @Gillibald maybe this is something to consider for Avalonia?

There's no before/after benchmarks unfortunately because I added them as an afterthought, however I took a sampling snapshot early in the process. As you can see, loading Unicode data is no longer a hotspot.

Before
image

After
image

I don't want to go crazy optimizing much further in this PR but will have a good scan through the sampling data to see if there's some low hanging fruit at another time.

CC @Sergio0694 @kant2002 @0xced


Gillibald:
Nice work @JimBobSquarePants

kant2002:
I'm not sure how do you get trie files, but I think it's better to keep source trie files in source control and document generation process. I understand that nobody want run file generation process on the build, but that process easy to forget, and maybe somebody would like to improve it, so better to have source files around. 
Or maybe I miss something in the changes ?

JimBobSquarePants:
Thanks @Gillibald your idea to create classes was a great one!

@kant2002 The trie files are created via the Generator project and simply contained a header and the parsed bytes. That project reads a bunch of Unicode spec data files and creates the output.

Here's the project in the `main` branch.
https://github.com/SixLabors/Fonts/tree/main/src/UnicodeTrieGenerator

If you look at `Generator.cs` in this PR we're simply switching out the file generation, instead opting to create classes containing the trie data. All the source files required to create those classes are in the solution.

We definitely do not need to build them every time, though the process is pretty quick since the spec files are stored locally. We actually won't need to build these again until Unicode 15 is released and that will require a dedicated PR to do so there's no opportunity to forget.

kant2002:
@JimBobSquarePants thanks for explanation. Yeah, now it make sense to me. Thanks you for caring about such niche workloads :smile:

Gillibald:
I will adapt your changes for Avalonia. Really appreciate your work.

tocsoft (approved):

DaZombieKiller:
I just thought it'd be worth pointing out that the `UnicodeTrie`-related code won't work on big endian systems (and will potentially fail or throw `DataMisalignedException` on systems where aligned access is required). Something more along the lines of this would solve both of those issues:
```cs
struct UnicodeTrieHeader
{
    public int HighStart;
    public uint ErrorValue;
    public int DataLength;
}

// ...

readonly uint[] data;
readonly int highStart;
readonly uint errorValue;

public UnicodeTrie(ReadOnlySpan<byte> rawData)
{
    // MemoryMarshal.Read instead of MemoryMarshal.Cast/Unsafe.As
    // because it performs an unaligned read and a bounds check.
    var header = MemoryMarshal.Read<UnicodeTrieHeader>(rawData);

    if (!BitConverter.IsLittleEndian)
    {
        header.HighStart  = BinaryPrimitives.ReverseEndianness(header.HighStart);
        header.ErrorValue = BinaryPrimitives.ReverseEndianness(header.ErrorValue);
        header.DataLength = BinaryPrimitives.ReverseEndianness(header.DataLength);
    }

    highStart  = header.HighStart;
    errorValue = header.ErrorValue;
    data       = new uint[header.DataLength / sizeof(uint)];

    // Avoid MemoryMarshal.Cast because we don't know if the rawData buffer is aligned
    // to uint, which would potentially throw DataMisalignedException if it weren't.
    rawData[Unsafe.SizeOf<UnicodeTrieHeader>()..].CopyTo(MemoryMarshal.AsBytes(data.AsSpan()));

    if (!BitConverter.IsLittleEndian)
    {
        for (int i = 0; i < data.Length; i++)
        {
            data[i] = BinaryPrimitives.ReverseEndianness(data[i]);
        }
    }
}
```

TechPizzaDev:
The data is coming from the `ReadOnlySpan<byte>` constants isn't it? That means that casting the bytes (which are most likely little-endian) into integers will break on big-endian.

JimBobSquarePants:
Ooh. You're right! The endianness definitely matters since it's a compiled constant. I'll update unless you want to create a PR @DaZombieKiller ?

DaZombieKiller:
@JimBobSquarePants Sure! Opened one here: https://github.com/SixLabors/Fonts/pull/284

PR #282, opened 2022-06-29
Add Bench mark for text render optimized TryGetGlyphMetricsAtOffset

Faolan-Rad:
### Description
This was me just finds Glyph Metrics  when it loads and updates rather then on every TryGetGlyphMetricsAtOffset which improves performance even more

### DIFF
--- a/src/SixLabors.Fonts/GlyphPositioningCollection.cs
+++ b/src/SixLabors.Fonts/GlyphPositioningCollection.cs
@@ -20,6 +20,11 @@ namespace SixLabors.Fonts
         private readonly List<GlyphPositioningData> glyphs = new();
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Faolan-Rad:
This is based off this pull request [Speed up Unicode Data lookups and remove reflection](https://github.com/SixLabors/Fonts/pull/281)

JimBobSquarePants:
Hi @Faolan-Rad thanks for your PR. 

To accept this I would have to see two things. 

1. Proof that this area is a processing hotspot and that it requires extra optimization.
2. Before/After benchmark results showing the improvement from the change. 

I would need some convincing to see the usefulness of this change. `GlyphPositioningCollection.TryGetGlyphMetricsAtOffset(...)` is only called during line breaking which involves a sequential run against the codepoints within the string. This means that a given offset is only requested one time which means any cache lookup will always return false.

PR #284, opened 2022-06-30
Fix potential endianness and alignment problems with Span-based UnicodeTrie constructor

DaZombieKiller:
### Prerequisites

### Description
This is a small PR making some changes to the recently-introduced `ReadOnlySpan<byte>` constructor for `UnicodeTrie`. Currently, the data provided to this constructor is constant data generated via the Roslyn `ROS<byte> = new byte[]` trick, which doesn't guarantee alignment on the data. The data is also stored in little-endian, which requires swapping for big-endian platforms.

* `UnicodeTrieHeader` is now a simple structure containing only fields, and it no longer has a `StructLayout` attribute. Its layout won't result in padding, so `Pack = 1` was removed.
* The `UnicodeTrie` constructor now reads `UnicodeTrieHeader` with `MemoryMarshal.Read`, to ensure there are no problems with alignment on platforms that require it (`MemoryMarshal.Read` performs an unaligned read.)
* The `rawData` span is no longer `Cast`ed to `uint` but rather the _destination_ `uint` array is casted to `byte`, for similar alignment-related reasons.
* When the platform is not little-endian, the constructor will now swap the endianness of the data before interpreting it.

### DIFF
--- a/src/SixLabors.Fonts/Unicode/UnicodeTrie.cs
+++ b/src/SixLabors.Fonts/Unicode/UnicodeTrie.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0.
 
 using System;
+using System.Buffers.Binary;
 using System.IO;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Haha... You beat me by seconds. Was just committing to a branch with your changes! 😁

JimBobSquarePants (approved):
Brilliant. Thanks! 👍

Issue #294, opened 2022-08-03
Hinting makes text looks like zalgo.


eX-pes:
### Description
I'm evaluating your library for my project. But when i tried to turn hinting on - text got broken: 
![_six](https://user-images.githubusercontent.com/71254535/182556775-48eaffe6-d1e3-4dfa-a07d-c3e329442adb.png)
Without hinting it's looks like this: 
![_six](https://user-images.githubusercontent.com/71254535/182556984-502221af-36eb-4618-8aae-98642f013cac.png)

### Steps to Reproduce
``` cs
public static void Main(string[] args)
{
    var fonts = new FontCollection();

    FontFamily font = fonts.Add(@"Fonts\OpenSans-Regular.ttf");

    using var img = new Image<Rgba32>(1024, 800);
    img.Mutate(x => x.Fill(Color.White));

    var str = "The quick brown fox jumps over the lazy dog. 0123456789";

    var off_y = 0f;

    for (int size = 5; size < 64; size++)
    {
        RenderText(font.CreateFont(size), str, img, off_y += size + 1);
    }

    using FileStream fs = File.Create("_six" + ".png");
    img.SaveAsPng(fs);
}

public static void RenderText(Font font, string text, Image<Rgba32> img, float y)
{
    string path = IOPath.GetInvalidFileNameChars().Aggregate(text, (x, c) => x.Replace($"{c}", "-"));
    string fullPath = IOPath.GetFullPath(IOPath.Combine("Output", IOPath.Combine(path)));

    IPathCollection shapes = TextBuilder.GenerateGlyphs(text, new TextOptions(font)
    {
        Origin = new Vector2(50f, y),
    });
    img.Mutate(x => x.Fill(Color.Black, shapes));
}
```

### System Configuration

- Fonts version: 1.0.0-beta18
- Other Six Labors packages and versions: SixLabors.ImageSharp.Drawing 1.0.0-beta14.8
- Environment (Operating system, version and so on): Win10
- .NET Framework version: netcore sdk 3.1.421

JimBobSquarePants:
See #293 

Also, please don’t change the template

JimBobSquarePants:
Nice test code btw. I’m going to use that while I finish fixing the issue

JimBobSquarePants:
@eX-pes can you please attach your version of the font?

jakerdy:
It was this one:
[...]

PR #295, opened 2022-08-11
Fix Font Hinting

### LINKED ISSUE #293: Small font rendering really unclear (blurry, missing pixels etc)
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of Fonts
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/Fonts/issues) and [closed](https://github.com/SixLabors/Fonts/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

Rendering small fonts is unreadable with alpha blending enabled, due to the really imprecise, blurry way they're being rendered:

ImageSharp (default antialiasing settings):
image

GDI+ (clear type):
image

This is with Segoe UI at 10 pixels. The vertical lines especially aren't clear, and with alpha blending enabled they just disappear. 

I also tried disabling antialiasing in the ImageSharp render, but that makes it even more obvious how imprecise it is:
image

I also tried hinting, which is supposed to help with this, but it doesn't seem to work at all:
image

For Open Sans it's even less readable, and hinting just produces long vertical lines:
image

Is there some set of options I can enable to make small font rendering clear?

### Steps to Reproduce
I have a sample repo that I used to generate those images at https://github.com/myblindy/font_test

### System Configuration

- Fonts version: 1.0.0-beta18
- Other Six Labors packages and versions: ImageSharp 2.1.3, ImageSharp.Drawing 1.0.0-beta15
- Environment (Operating system, version and so on): Windows 10
- .NET Framework version: .Net 6
- Additional information: N/A

JimBobSquarePants:
### Prerequisites

### Description

Fixes #293 and #294

We were passing the multiplied pixel size when we shouldn't have. Some pre-cleartype fonts don't seem to work all that well so I've internally disabled hinting for them for now.


src/SixLabors.Fonts/StreamFontMetrics.TrueType.cs
```diff
@@ -21,11 +21,31 @@ namespace SixLabors.Fonts
...
+        /// These common fonts can be hinted by the engine only in XY mode and only if we do not scale the x-dimension.
+        /// even then the output is worse than the non-hinted version due to thinning of certain strokes.
+        /// <see href="http://agg.sourceforge.net/antigrain.com/research/font_rasterization/"/>.
+        /// </para>
+        /// <para>
+        /// We'll sacrifice hinting for these fonts to allow the engine to work better for modern fonts and avoid potentially raised issues.
+        /// </para>
+        /// </summary>
+        private static readonly string[] BadHintList = { "Arial", "Times New Roman", "Tahoma", "Palatino Linotype", "Segoe UI" };
+        private bool? shouldHint;
```
JimBobSquarePants, line 36: There's a LOT of knowledge in this document, unfortunately I couldn't find a way to get these fonts to hint well. They're all older ones that have known issues with hinting so to save effort I've disabled hinting of them for now.

http://rastertragedy.com/RTRCh4.htm#Sec1

Every single modern font I tested seems to work really well.

tocsoft: is there any metadata in the font we can use to determine if we should enable hinting? some created date maybe or so versioning flag difference? anything to save just hard coding a set of font names.

JimBobSquarePants: Unfortunately not. 😔

There's some guidance for how to actually mitigate the issues I saw but I spent a few days and couldn't translate it into code. Nor could I find reference in FreeType that provided a complete workaround matching that guidance.

> Accordingly I was looking for an approach that combines the advantages of overscaling before constraint application with the “safety” of overscaling after constraint application. To do so I “jury rigged” the rasterizer to dynamically re-interpret all rounding instructions to mean “round to sample boundary” if the rendering method is ClearType and if the direction of the projection vector p is not parallel to the long side of the LCD sub-pixels, else to keep the meaning “round to pixel boundary” as before:
> 
> - Apply contextually interpreted constraints (“hints”) to outlines
> - Overscale constrained outlines in x-direction by 6x
> - Sample overscaled constrained outlines into overscaled bitmaps
> - Apply anti-aliasing filter to downsample overscaled bitmaps
> 
> Semantically, the two interpretations of rounding instructions are most similar. In fact, if you consider full-pixel bi-level rendering (colloquially “black-and-white” rendering) as “oversampling” at the rate of one sample per pixel, then the two interpretations are identical. Practically, this approach eliminates the “explosions” and it overcomes the absence of a TrueType instruction that would permit “finer grained” rounding than to the nearest 1/2 of a pixel (cf [4.3.0](http://rastertragedy.com/RTRCh4.htm#Sec30)).

src/SixLabors.Fonts/Tables/TrueType/Hinting/TrueTypeInterpreter.cs
```diff
@@ -1398,7 +1399,7 @@ private void ShiftPoints(Vector2 displacement)
             state.Loop = 1;
         }
 
-        private void MovePoint(Zone zone, int index, float distance)
+        private void MovePoint(Zone zone, int index, float distance, [CallerMemberName] string member = null)
```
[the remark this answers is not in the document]
JimBobSquarePants: Ah yes. That was me trying to figure out where exactly a glyph  explosion was getting triggered.

src/SixLabors.Fonts/StreamFontMetrics.TrueType.cs
JimBobSquarePants: I do wonder whether we just just try tagging the chap who built the original interpreter to see if he knows how?

src/SixLabors.Fonts/Buffer{T}.cs
```diff
@@ -26,21 +27,37 @@ public Buffer(int length)
...
+        /// </summary>
+        /// <returns>The <see cref="ArraySlice{T}"/>.</returns>
+        public ArraySlice<T> DangerousGetSlice()
+        {
+            if (this.buffer is null)
+            {
+                ThrowObjectDisposedException();
+            }
+
+            return new ArraySlice<T>(Unsafe.As<T[]>(this.buffer), 0, this.length);
```
JimBobSquarePants, line 50: This is dangerous. Will look for a workaround.

src/SixLabors.Fonts/StreamFontMetrics.TrueType.cs
JimBobSquarePants: I hate to do this and apologies in advance but @MikePopoloski would you have any insight into how to implement some of these workarounds?

MikePopoloski: Oh hello. No sorry, I haven't thought much about font rendering in like 7 years. Glad someone is making use of the interpreter though, it's a neat bit of code and there aren't (or at least weren't when I built it) too many open source implementations of a TT interpreter.

JimBobSquarePants: Thanks for replying! That was a bit of a hit and hope tagging you like that. 

I think there's very few people in the world who actually understand the TT interpreter. I've yet to find an implementation other than yours in the .NET ecosystem and elsewhere there just appears to be using Freetype.

I'll persevere though and continue to do my own research. Fongers crossed I can figure it all out. It's an interesting challenge!

src/SixLabors.Fonts/Buffer{T}.cs
JimBobSquarePants: @antonfirsov Do you have any ideas here?

I never let anything access the buffer outside of the range, but it still doesn't feel right that I have this unsafe cast array that returns the length in the number of bytes.

JimBobSquarePants: @antonfirsov sorry to bump. I'm just out of ideas without rewriting the interpreter to work with spans

JimBobSquarePants: Decided to take the perf hit and go safe for now.

JimBobSquarePants:
@tocsoft I've cracked it! 🎉

I had a good dig through the FreeType interpreter code and accompanying docs and figured out what they did to get their interpreter working well with legacy fonts. Basically, they never move anything horizontally when hinting at all which fixes most issues then there's a couple of additional tweaks for legacy fonts. No need for lists or targeted hacks - it just works!

Here's Tahoma (_This was really bad before with odd offsetting of individual glyphs._)
 
`HintingMode.HintXY`
_six_tahoma_hintxy_

`HintingMode.HintY`
_six_tahoma_hinty_

`HintingMode.None`
_six_tahoma_none_

`HintingMode.HintXY` vs `HintingMode.None` (_You can clearly see the hinting benefit in the 'o' here_)
image

And here's Open Sans

`HintingMode.HintXY` (_Both hint modes Y and XY produce identical results with this font_)
_six_open_sans_hintxy_

`HintingMode.None`
_six_open_sans_none_

`HintingMode.HintXY` vs `HintingMode.None`  (_You can clearly see the hinting benefit in the 'o' here_)
image

For most fonts there really miniscule difference between `HintingMode.HintXY` and `HintingMode.HintY`. There are some notable improvements in Times New Roman when using `HintingMode.HintY` in that the over-thinning of stems does not happen. I'm thinking of reworking the enum to rename `HintingMode.HintY` as `HintingMode.Compatibility` moving it to the third position as there's additional overheads (_scaling of control point x-values before and after hinting_) which, while SIMD optimized, you don't really want people using as a first-choice approach.

What do you think?

image

tocsoft (dismissed):
This looks fantastic, congrats on getting it working without the black list.

tocsoft (approved):

Issue #296, opened 2022-08-18
Woff font loading broken in latest beta - Could not read compressed data! 


mj2015:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of Fonts
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/Fonts/issues) and [closed](https://github.com/SixLabors/Fonts/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
I have code that loads WOFF fonts from files on disk, then renders them to a bitmap. This worked fine in version 1. I updated to the latest Nuget betas, and now I get an error when calling `TextMeasurer.Measure` with a reference to the font. The error is `Could not read compressed data! Expected bytes: 35776, bytes read: 14988` which implies to me that the font loading has failed. Full stack is:

`   at SixLabors.Fonts.Tables.Woff.WoffTableHeader.CreateReader(Stream stream)
   at SixLabors.Fonts.FontReader.GetReaderAtTablePosition(String tableName)
   at SixLabors.Fonts.Tables.TrueType.Glyphs.GlyphTable.Load(FontReader reader)
   at SixLabors.Fonts.Tables.TableLoader.Load[TTable](FontReader reader)
   at SixLabors.Fonts.FontReader.TryGetTable[TTableType]()
   at SixLabors.Fonts.FontReader.GetTable[TTableType]()
   at SixLabors.Fonts.StreamFontMetrics.LoadTrueTypeFont(FontReader reader)
   at SixLabors.Fonts.StreamFontMetrics.LoadFont(String path, Int64 offset)
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at SixLabors.Fonts.FileFontMetrics.TryGetGlyphId(CodePoint codePoint, Nullable`1 nextCodePoint, UInt16& glyphId, Boolean& skipNextCodePoint)
[...]

PR #297, opened 2022-08-19
Fix reading compressed WOFF data

brianpopow:
### Prerequisites

### Description

This PR fixes reading compressed woff data: Read data in a loop. The first read does not ensure all data is read.

Fix issue #296

Note: I cant use the font file provided in #296 for a unit test. Not sure howto make a test without a sample file.

### DIFF
--- a/src/SixLabors.Fonts/Tables/Woff/WoffTableHeader.cs
+++ b/src/SixLabors.Fonts/Tables/Woff/WoffTableHeader.cs
@@ -26,15 +26,17 @@ namespace SixLabors.Fonts.Tables.Woff
             using var compressedStream = new IO.ZlibInflateStream(stream);
             byte[] uncompressedBytes = new byte[this.Length];
             int totalBytesRead = 0;
+            int bytesLeftToRead = uncompressedBytes.Length;
             while (totalBytesRead < this.Length)
             {
-                int bytesRead = compressedStream.Read(uncompressedBytes, 0, uncompressedBytes.Length);
+                int bytesRead = compressedStream.Read(uncompressedBytes, totalBytesRead, bytesLeftToRead);
                 if (bytesRead <= 0)
                 {
                     throw new InvalidFontFileException($"Could not read compressed data! Expected bytes: {this.Length}, bytes read: {totalBytesRead}");
                 }
 
                 totalBytesRead += bytesRead;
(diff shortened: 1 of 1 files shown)


tocsoft (approved):

dlemstra:
This probably deserved a unit test? With a custom stream that reads the bytes in blocks?

brianpopow:
> This probably deserved a unit test? With a custom stream that reads the bytes in blocks?

@dlemstra I agree that we should have tests for this, but I dont think a custom stream would be enough as this issue is specifically related to a change in how .Net 6.0 handles ZlibInflateStream.Read.

We currently do not run the test for .Net6.0 and I think we should. In fact the issue here is already covered by tests, we just need to target .Net 6.0 with the tests (I have tried that).

@JimBobSquarePants Is there a reason we are still targeting the old frameworks? Like `netcoreapp3.1;netcoreapp2.1;net472`. Should we drop those and replace them with .Net6.0?

JimBobSquarePants:
I would absolutely love to drop those targets but that means dropping them for ImageSharp.Drawing also. 

What do we think @SixLabors/core ?

brianpopow:
> I would absolutely love to drop those targets but that means dropping them for ImageSharp.Drawing also.
> 
> What do we think @SixLabors/core ?

My opinion is, we should drop support for the old frameworks. I dont see why we should support them, when even microsoft does not do that.

mj2015:
Can I ask how often releases are done? If I understand correctly, there is a fix in the code for this issue, but there hasn't been a release done since the beginning of July. I'd really like access to the fix.

As for maintaining backward compatibility, Core 3.1 is going out of support, and I'm only using .Net6 in any new work. Given the library is pretty much for newer projects, I can't see why you'd keep support. Thanks.

JimBobSquarePants:
We're trying to limit Nuget beta releases while we work on an RC release, however, every build against our main branch is published on our MyGet feed.

mj2015:
Thank you for that - I can confirm that the fix works well here. 

Also a further thought on the dropping of old frameworks - it is based on my thinking that this project is primarily aimed at developers who are wanting multi-platform operation, and that is .Net 6 right now. If you are on framework, you'll use the Windows specifics, and if you are on an old Core version, it isn't too hard to update.
Thanks again.

PR #308, opened 2022-11-23
Add exposed function to retrieve all unicode codepoints defined in fonts

CptWesley:
### Prerequisites

- [X] I have written a descriptive pull-request title
- [X] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/Fonts/pulls) open
- [X] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [X] I have provided test coverage for my change (where applicable)

### Description

Fixes #307

As described in the suggestion, exposing this information can make it easier for users to retrieve some extra information about what code points and glyphs are available in a font and how they are being mapped. Whether or not this is the right strategy to allow these queries is of course up for debate.

Changes:
- Added `GetAvailableCodePoints` function to `Font` type which retrieves all defined code points in that particular font.
- Made `GlyphId` field in `GlyphMetrics` public.

### DIFF
--- a/src/SixLabors.Fonts/Font.cs
+++ b/src/SixLabors.Fonts/Font.cs
@@ -167,6 +167,13 @@ namespace SixLabors.Fonts
             }
         }
 
+        /// <summary>
+        /// Gets the unicode codepoints for which a glyph exists in the font.
+        /// </summary>
+        /// <returns>An enumerable containing all available codepoints.</returns>
+        public IEnumerable<CodePoint> GetAvailableCodePoints()
+            => this.FontMetrics.GetAvailableCodePoints();
+
         private string LoadFontName()
             => this.metrics.Value?.Description.FontName(this.Family.Culture) ?? string.Empty;
 
--- a/src/SixLabors.Fonts/Font.cs
+++ b/src/SixLabors.Fonts/Font.cs
@@ -167,13 +167,6 @@ namespace SixLabors.Fonts
             }
         }
 
-        /// <summary>
-        /// Gets the unicode codepoints for which a glyph exists in the font.
-        /// </summary>
-        /// <returns>An enumerable containing all available codepoints.</returns>
-        public IEnumerable<CodePoint> GetAvailableCodePoints()
-            => this.FontMetrics.GetAvailableCodePoints();
-
         private string LoadFontName()
             => this.metrics.Value?.Description.FontName(this.Family.Culture) ?? string.Empty;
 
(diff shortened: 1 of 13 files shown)


src/SixLabors.Fonts/Font.cs
```diff
@@ -167,6 +167,13 @@ public IEnumerable<Glyph> GetGlyphs(CodePoint codePoint, TextAttributes textAttr
             }
         }
 
+        /// <summary>
+        /// Gets the unicode codepoints for which a glyph exists in the font.
+        /// </summary>
+        /// <returns>An enumerable containing all available codepoints.</returns>
+        public IEnumerable<CodePoint> GetAvailableCodePoints()
```
JimBobSquarePants, line 174: This should probably be IReadOnlyCollection<CodePoint> since we want it to be a one-time enumeration.

tests/SixLabors.Fonts.Tests/FontCodePointsTests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace SixLabors.Fonts.Tests
+{
+    public class FontCodePointsTests
+    {
+        [Fact]
+        public void TtfTest()
+        {
+            var collection = new FontCollection();
+            FontFamily family = collection.Add(TestFonts.SimpleFontFile);
```
JimBobSquarePants, line 17: I think we should run these tests using a real font.

src/SixLabors.Fonts/Tables/General/CMap/Format14SubTable.cs
```diff
@@ -138,6 +139,9 @@ public override bool TryGetGlyphId(CodePoint codePoint, out ushort glyphId)
             return false;
         }
 
+        public override IEnumerable<int> GetAvailableCodePoints()
+            => Array.Empty<int>();
```
JimBobSquarePants, line 143: Shouldn't we return the variationSelectors keys here?

JimBobSquarePants (changes requested):
Hi there. Thanks for making an effort here. This looks pretty good!

I think we should implement a few performance tweaks though and I'm unsure about the expected behavior.

JimBobSquarePants:
I haven't forgotten about this. I'm just doing some research to ensure we can implement the correct thing.

CptWesley:
My apologies for no longer replying due to the holiday season. I have not incorporated the requested changes due to the fact that it seems this might not be the right approach for solving this due to what you mentioned and what was mentioned in #307. My personal experience with font rendering is very minimal, so I'm not sure what the right approach would be, but if I can be of any help, please let me know.

JimBobSquarePants:
Hi @CptWesley I did some research, and I cannot find a definitive answer regarding variations. As such I believe it's best to leave as-is. 

The rest of your code is ace. All I changed was the review comments re optimization. Thanks!

JimBobSquarePants (approved):

CptWesley:
@JimBobSquarePants Thank you very much!

PR #326, opened 2023-03-16
Fix rendering font size and overline position.

JimBobSquarePants:
### Prerequisites

### Description

Fixes a regression where the current run font size was not being used for rendering. Also fixed overline position to match CSS


JimBobSquarePants:
I've already directly tested this locally against https://github.com/SixLabors/ImageSharp.Drawing/pull/211

PR #328, opened 2023-04-19
Add apply test and fix apply

stefannikolei:
fix dfa apply and add the test

### DIFF
--- a/src/SixLabors.Fonts/Unicode/Dfa/StateMachine.cs
+++ b/src/SixLabors.Fonts/Unicode/Dfa/StateMachine.cs
@@ -112,7 +112,8 @@ namespace SixLabors.Fonts.Unicode.Dfa
                 {
                     if (actions.TryGetValue(tag, out Action<int, int, int[]>? action))
                     {
-                        action(match.StartIndex, match.EndIndex + 1 - match.StartIndex, input);
+                        // can be optimized... range could not be used because of the oooold targets.
+                        action(match.StartIndex, match.EndIndex, input.Skip(match.StartIndex).Take(match.EndIndex + 1 - match.StartIndex).ToArray());
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


src/SixLabors.Fonts/Unicode/Dfa/StateMachine.cs
```diff
@@ -112,7 +112,8 @@ public void Apply(int[] input, Dictionary<string, Action<int, int, int[]>> actio
                 {
                     if (actions.TryGetValue(tag, out Action<int, int, int[]>? action))
                     {
-                        action(match.StartIndex, match.EndIndex + 1 - match.StartIndex, input);
+                        // can be optimized... range could not be used because of the oooold targets.
+                        action(match.StartIndex, match.EndIndex, input.Skip(match.StartIndex).Take(match.EndIndex + 1 - match.StartIndex).ToArray());
```
JimBobSquarePants, line 116: Yeah. Annoying eh? I was thinking we might be able to use a span inside the caller action and avoid the allocation here

tests/SixLabors.Fonts.Tests/SixLabors.Fonts.Tests.csproj
```diff
@@ -8,6 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="FluentAssertions" Version="6.10.0" />
```
JimBobSquarePants, line 11: I don’t use this library and don’t want to introduce it as I’m not a fan of the style

JimBobSquarePants (commented):
Thanks for looking at this 👍

JimBobSquarePants:
I had some time on this and pushed my own version using the `ArraySlice<T>` type to avoid the unnecessary allocation and pre-slice the output. Thanks for your help anyway!!@

PR #332, opened 2023-05-23
Rewrite Text Layout Engine

JimBobSquarePants:
### Prerequisites

### Description

When rendering single line text both the orientation and layout of text within the box was incorrectly calculated for single line text, ignoring the set wrapping length. This PR fixes both algorithms to respect that value.

https://github.com/SixLabors/Fonts/issues/335

UPDATE. When writing tests I realized we had multiple issues with our layout engine so I rewrote it to match exactly the output produced by browsers.

I've also added the ability to measure 3 sets of bounds per textbox and for individual glyphs.

- Advance : The line-height and advanced width
- Size : The minimum size to render the box or glyphs. Leading advances are included but following are not.
- Bounds : The exact bounds of the box or glyphs.

In the following images below, you can see various boxes indicating the results of those measurements. These work for all layout options.

issue_275_Top_Left_Start - Copy
issue_275_Top_Left_Start
issue_275_Top_Left_Start - Copy (2)

### DIFF
--- a/src/SixLabors.Fonts/FontFamily.cs
+++ b/src/SixLabors.Fonts/FontFamily.cs
@@ -155,7 +155,7 @@ namespace SixLabors.Fonts
         /// <see langword="true"/> if the <see cref="FontFamily"/> contains font metrics
         /// with the specified name; otherwise, <see langword="false"/>.
         /// </returns>
-        internal bool TryGetMetrics(FontStyle style, [NotNullWhen(true)] out FontMetrics? metrics)
+        public bool TryGetMetrics(FontStyle style, [NotNullWhen(true)] out FontMetrics? metrics)
         {
             if (this == default)
             {
--- a/src/SixLabors.Fonts/GlyphRendererParameters.cs
+++ b/src/SixLabors.Fonts/GlyphRendererParameters.cs
@@ -19,7 +19,7 @@ namespace SixLabors.Fonts
             TextRun textRun,
             float pointSize,
             float dpi,
-            LayoutMode layoutMode)
+            GlyphLayoutMode layoutMode)
         {
             this.Font = metrics.FontMetrics.Description.FontNameInvariantCulture?.ToUpper() ?? string.Empty;
             this.FontStyle = metrics.FontMetrics.Description.Style;
@@ -76,7 +76,7 @@ namespace SixLabors.Fonts
         /// <summary>
         /// Gets the layout mode applied to the glyph.
         /// </summary>
-        public LayoutMode LayoutMode { get; }
+        public GlyphLayoutMode LayoutMode { get; }
 
         /// <summary>
         /// Gets the text run that this glyph belongs to.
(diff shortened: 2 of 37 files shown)


tocsoft (commented):
I feel we are missing a test case for this.

JimBobSquarePants:
> I feel we are missing a test case for this.

Agreed. Very late here now so I'll write some tomorrow.

JimBobSquarePants:
@tocsoft While writing tests I found several issues with our bounds measuring functionality. I'm expanding on this PR to fix these issues and add appropriate tests.

JimBobSquarePants:
@tocsoft I still need to add some additional tests but have completed the rewrite of the layout engine. It's a lot simpler now and much more accurate. 

I've done a lot of local visual testing using Drawing and we now exactly match browser text layout implementations.

src/SixLabors.Fonts/FontFamily.cs
```diff
@@ -155,7 +155,7 @@ public bool TryGetPaths(out IEnumerable<string> paths)
         /// <see langword="true"/> if the <see cref="FontFamily"/> contains font metrics
         /// with the specified name; otherwise, <see langword="false"/>.
         /// </returns>
-        internal bool TryGetMetrics(FontStyle style, [NotNullWhen(true)] out FontMetrics? metrics)
+        public bool TryGetMetrics(FontStyle style, [NotNullWhen(true)] out FontMetrics? metrics)
```
JimBobSquarePants, line 158: This is too useful to hide.

src/SixLabors.Fonts/GlyphRendererParameters.cs
```diff
@@ -19,7 +19,7 @@ namespace SixLabors.Fonts
             TextRun textRun,
             float pointSize,
             float dpi,
-            LayoutMode layoutMode)
+            GlyphLayoutMode layoutMode)
```
JimBobSquarePants, line 22: I've updated Drawing locally to cater for this.

tests/SixLabors.Fonts.Tests/Tables/AdvancedTypographic/GPos/GPosTableTests.cs
```diff
@@ -21,8 +21,8 @@ public void SingleAdjustmentPositioning_Format1_Works()
             int[] expectedGlyphIndices = { 20, 22 };
             FontRectangle[] expectedFontRectangles =
             {
-                new(275, 1811, 825, 719),
-                new(1622, 1810, 978, 704),
+                new(275, 1085.9999F, 825, 719),
```
JimBobSquarePants, line 24: An unfortunate side effect of using single precision floating point.

src/SixLabors.Fonts/TextOptions.cs
```diff
@@ -88,20 +87,12 @@ public float Dpi
...
-            set
-            {
-                Guard.MustBeGreaterThanOrEqualTo(value, 0, nameof(this.TabWidth));
-                this.tabWidth = value;
-            }
-        }
+        /// <remarks>
+        /// If value is -1 then the font default tab width is used.
+        /// </remarks>
+        public float TabWidth { get; set; } = -1F;
```
JimBobSquarePants, line 95: Our original implementation wasn't accurate and was actually multiplying the tab advance. This now really uses spaces.

tocsoft (approved):

PR #340, opened 2023-06-29
Fix DFKai-SB and KaiU font rendering

### LINKED ISSUE #337: DrawText generates text with DFKai-SB or KaiU fonts differently than expected
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp.Drawing
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp.Drawing/issues) and [closed](https://github.com/SixLabors/ImageSharp.Drawing/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

using chinese fonts on DrawText I got the error pic

### System Configuration

- ImageSharp.Drawing version:1.0.0-beta15
- Other ImageSharp packages and versions: 3.0.1
- .NET Framework version: .net 7
`       

           var font = SystemFonts.CreateFont(fontName, fontsize);
            var color = new Rgba32((byte)colorR, (byte)colorG, (byte)colorB);

            var image = new Image<Rgba32>(1, 1);
            var textSize = TextMeasurer.Measure(txt, new TextOptions(font));
            image.Mutate(x => x.Resize(new Size((int)textSize.Width, (int)textSize.Height)));
            image.Mutate(x => x.DrawText(txt, font, color, PointF.Empty));

            using (var stream = new MemoryStream())
            {
                image.Save(stream, new PngEncoder());
                return stream.ToArray();
            }
`

JimBobSquarePants:
### Prerequisites

### Description

Fixes #337 

There's a lot going on here. 

The font in question requires hinting in order to be displayed properly (_Thanks @behdad !!_) but we were not loading the instructions for composite glyphs from the font. As such, I ended up doing a significant refactoring (_massively simplifying_) of our glyph loading mechanism to make it possible to do so. There is a lot less cloning going on now so far fewer allocations will take place.

 TrueTypeInterpreter changes are based on Freetype's implementation and all tests visual and unit remain unaffected.

Here's the correctly rendered font.

sb

Questions.

We do not hint by default as it creates overhead and legacy fonts such as Arial can have minor errors at some font sizes yet this font specifically requires hinting. Should we do as FreeType do and explicitly force hinting if the font is in a known list?

https://github.com/behdad/freetype/blob/08f66322e32cc882733dfae408e040f5057ee2ac/src/truetype/ttobjs.c#L181


tocsoft (approved):
code as it stand look good to me.

The question on overriding the hinting I feel is a little thorny. It feels wrong each font shaper/renderer having to maintain magic lists of special case fonts all over the place... having a global resource tracking these, arguably broken, fonts might be a useful tool to maintain (something that all the libraries can benefit from).. or users can see if they are using specific fonts need specific override settings.

We should then provide a mechanism at the `FontCollection` level to declare these sorts of override (forcing hinting on/off etc) forcing some specific shapings in other situations maybe too.

These sorts of font issues should be a thing that developers are made aware of, which is why I think an online tool/resource tracking these things and what the overrides shoud be is real answer here. It will end up being a never ending loop of 'I have a boken font, please add the overrides to lib x & lib y etc'.... it should be something that a user of `Font`s can self discover and apply the fix in thier code themselves... which will be espcially important for those that are using it in applications where their users are supplying thier own fonts.

If the tracker/online tool is built right it could also be used as a source of metadata needed to auto produce a nuget that can apply all its defaults even... but built/released independently to `Font` as and when fonts with issues are discovered and configured with their fixes.

JimBobSquarePants:
> code as it stand look good to me.
> 
> The question on overriding the hinting I feel is a little thorny. It feels wrong each font shaper/renderer having to maintain magic lists of special case fonts all over the place... having a global resource tracking these, arguably broken, fonts might be a useful tool to maintain (something that all the libraries can benefit from).. or users can see if they are using specific fonts need specific override settings.
> 
> We should then provide a mechanism at the `FontCollection` level to declare these sorts of override (forcing hinting on/off etc) forcing some specific shapings in other situations maybe too.
> 
> These sorts of font issues should be a thing that developers are made aware of, which is why I think an online tool/resource tracking these things and what the overrides shoud be is real answer here. It will end up being a never ending loop of 'I have a boken font, please add the overrides to lib x & lib y etc'.... it should be something that a user of `Font`s can self discover and apply the fix in thier code themselves... which will be espcially important for those that are using it in applications where their users are supplying thier own fonts.
> 
> If the tracker/online tool is built right it could also be used as a source of metadata needed to auto produce a nuget that can apply all its defaults even... but built/released independently to `Font` as and when fonts with issues are discovered and configured with their fixes.

Yep. Greate initiative!!

I've created https://github.com/SixLabors/docs/issues/60 for now so we at least document them.

PR #348, opened 2023-08-12
Simplify dependencies

jairbubbles:
[System.Memory](https://www.nuget.org/packages/System.Memory/) depends on the 3 other packages :

image

So they don't need to be explicitely referenced.


JimBobSquarePants:
Thanks for this. For v2 we're actually dropping those earlier frameworks, so we don't need to simplify the dependencies. I've repointed your PR at the v1 branch though so should we release any security updates your change will be included.

PR #371, opened 2023-12-19
Platform-independent way to build path for fonts into the sample. Now…

snechaev:
… should work on non-Windows systems.

### Prerequisites

### Description

The `DrawWithImageSharp` were crashing on macos/net6 with the `System.IO.FileNotFoundException` while adding a fonts from files with the paths like the `/Users/user/work/Fonts/artifacts/bin/samples/DrawWithImageSharp/Debug/net6.0/Fonts\\SixLaborsSampleAB.ttf` (note the `\\` delimiter before the file name). This pull request uses the `Path.Combine` to build the paths, so the correct platform-specific path delimiters are used.

### DIFF
--- a/samples/DrawWithImageSharp/Program.cs
+++ b/samples/DrawWithImageSharp/Program.cs
@@ -21,18 +21,18 @@ public static class Program
     {
         var fonts = new FontCollection();
         var woffFonts = new FontCollection();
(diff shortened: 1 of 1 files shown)


JimBobSquarePants (approved):
Thanks for this!

PR #380, opened 2024-02-08
Fix Woff2 bounds decompression and size measurement calculations.

JimBobSquarePants:
### Prerequisites

### Description

Fixes #375 #373 #367

We were calculating the bounds bitmap length calculation incorrectly which led to the wrong bounds being read.

Further investigation into determining the correct bounds led to the discovery that the bounds calculation could be simplified and that the size was being incorrectly calculated. 

I've added some additional visual tooling in the samples project that helps to verify measurement results. This helped me to discover that we were not rendering the results of some ligature substitutions if the remaining glyph code point was non-rendering. The output is pretty cool. 

I also fixed the line breaking algorithm to include a missed breaking opportunity and exactly match browser output.


src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -552,7 +556,9 @@ private static IReadOnlyList<GlyphLayout> LayoutText(TextBox textBox, TextOption
                     advanceX,
                     data.ScaledAdvance,
                     GlyphLayoutMode.Vertical,
-                    i == 0));
+                    i == 0 && j == 0));
```
JimBobSquarePants, line 559: Prevents individual metrics within a composite glyph adding to the advance.

src/SixLabors.Fonts/GlyphSubstitutionCollection.cs
```diff
@@ -276,15 +291,30 @@ public void Replace(int index, int count, ushort glyphId)
     {
         // Remove the glyphs at each index.
         int codePointCount = 0;
+        CodePoint codePoint = default;
         for (int i = count; i > 0; i--)
         {
             int match = index + i;
             codePointCount += this.glyphs[match].Data.CodePointCount;
+            CodePoint currentCodePoint = this.glyphs[match].Data.CodePoint;
+            if (!UnicodeUtility.IsDefaultIgnorableCodePoint((uint)codePoint.Value) || UnicodeUtility.ShouldRenderWhiteSpaceOnly(codePoint))
```
JimBobSquarePants, line 300: When substituting a ligation, we need to ensure that the remaining codepoint is renderable.

src/SixLabors.Fonts/TextMeasurer.cs
```diff
@@ -208,14 +208,8 @@ internal static FontRectangle GetAdvance(IReadOnlyList<GlyphLayout> glyphLayouts
...
     {
-        if (glyphLayouts.Count == 0)
-        {
-            return FontRectangle.Empty;
-        }
-
-        Vector2 topLeft = glyphLayouts[0].BoxLocation * dpi;
         FontRectangle bounds = GetBounds(glyphLayouts, dpi);
-        return new FontRectangle(0, 0, MathF.Ceiling(bounds.Right - topLeft.X), MathF.Ceiling(bounds.Bottom - topLeft.Y));
+        return new FontRectangle(0, 0, MathF.Ceiling(bounds.Width), MathF.Ceiling(bounds.Height));
```
JimBobSquarePants, line 212: We know the bounds width/height here so should use it directly.

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -1166,35 +1184,6 @@ private static void SubstituteBidiMirrors(FontMetrics fontMetrics, GlyphSubstitu
                         : metric.FontMetrics.VerticalMetrics;
                     float ascender = metricsHeader.Ascender * scaleY;
 
-                    // Adjust ascender for glyphs with a negative tsb. e.g. emoji to prevent cutoff.
-                    if (!CodePoint.IsWhiteSpace(codePoint))
```
JimBobSquarePants, line 1170: This was a hack based upon incorrect assumptions. The proper fix is to use a larger line spacing to meet the font design.

tocsoft (approved):

Issue #393, opened 2024-03-15
Issue 367 introduced lingering whitespaces


batwomankt:
### Discussed in https://github.com/SixLabors/Fonts/discussions/392

<div type='discussions-op-text'>

<sup>Originally posted by **batwomankt** March 15, 2024</sup>
Before the fix for issue 367 was merged, if the character at the end of a split for word wrap was whitespace, it would be trimmed.  However, with the 367 fix introduced, that particular bit of trimming is missed in `SplitAt` and images have lingering whitespaces at the end of the line that are not trimmed properly.

After some fiddling, I have a suggestion on how to have both the desired line split behavior and whitespace trimming.  This still passes the 3 line count desired by the opener of issue 367, while also removing any trailing whitespaces from text lines!

In the `TextLine` class, add a method something like:

```
private void TrimTrailingWhitespaceAndRecalculateMetrics()
{
    int index = this.data.Count;
    while (index > 0)
    {
        if (!CodePoint.IsWhiteSpace(this.data[index - 1].CodePoint))
        {
            break;
        }

        index--;
    }

    if (index < this.data.Count)
    {
        this.data.RemoveRange(index, this.data.Count - index);
    }

    // Lastly recalculate this line metrics.
    float advance = 0;
    float ascender = 0;
    float descender = 0;
    float lineHeight = 0;
    foreach (GlyphLayoutData glyph in this.data)
    {
        advance += glyph.ScaledAdvance;
        ascender = MathF.Max(ascender, glyph.ScaledAscender);
        descender = MathF.Max(descender, glyph.ScaledDescender);
        lineHeight = MathF.Max(lineHeight, glyph.ScaledLineHeight);
    }

    this.ScaledLineAdvance = advance;
    this.ScaledMaxAscender = ascender;
    this.ScaledMaxDescender = descender;
    this.ScaledMaxLineHeight = lineHeight;
}
```

and have the `SplitAt` function adjusted to something like:
```
        public TextLine SplitAt(LineBreak lineBreak, bool keepAll)
        {
            int index = this.data.Count;
            GlyphLayoutData glyphWrap = default;
            while (index > 0)
            {
                glyphWrap = this.data[--index];

                if (glyphWrap.Offset == lineBreak.PositionWrap)
                {
                    break;
                }
            }

            if (index == 0)
            {
                // Now trim trailing whitespace from this line.
                this.TrimTrailingWhitespaceAndRecalculateMetrics();
                return this;
            }

[...]

PR #395, opened 2024-03-15
trim whitespace from EOLs when line split is found

batwomankt:
### Prerequisites

### Description
In `SplitAt`, when a `TextLine` is split at exact length match - the loop will no longer grab and trim the whitespace on the next iteration because it exits early (Issue 367 fix).  Instead, added a whitespace trim in `SplitAt` when the exact length match is being examined.

See [Issue 393](https://github.com/SixLabors/Fonts/issues/393) for a visual representation.  More detail can be provided upon request.

### DIFF
--- a/src/SixLabors.Fonts/TextLayout.cs
+++ b/src/SixLabors.Fonts/TextLayout.cs
@@ -1358,7 +1358,14 @@ internal static class TextLayout
             this.data.RemoveRange(index, this.data.Count - index);
 
             // Now trim trailing whitespace from this line.
-            index = this.data.Count;
+            this.TrimTrailingWhitespaceAndRecalculateMetrics();
+
+            return result;
+        }
+
+        private void TrimTrailingWhitespaceAndRecalculateMetrics()
+        {
+            int index = this.data.Count;
             while (index > 0)
             {
                 if (!CodePoint.IsWhiteSpace(this.data[index - 1].CodePoint))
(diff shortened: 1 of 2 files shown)


tests/SixLabors.Fonts.Tests/Issues/Issues_367.cs
```diff
@@ -23,7 +23,7 @@ public void ShouldMatchBrowserBreak()
         Assert.Equal(3, lineCount);
 
         FontRectangle advance = TextMeasurer.MeasureAdvance(text, options);
-        Assert.Equal(365, advance.Width);
+        Assert.Equal(355, advance.Width);
```
JimBobSquarePants, line 26: The code you've added doesn't change the behavior. You've just moved the trimming function to a new method. Are there more changes to come?

src/SixLabors.Fonts/TextLayout.cs
```diff
@@ -1328,6 +1328,9 @@ public TextLine SplitAt(LineBreak lineBreak, bool keepAll)
 
                 if (index == 0)
                 {
+                    // Now trim trailing whitespace from this line in the case of an exact
+                    // length line break (non CJK)
+                    this.TrimTrailingWhitespaceAndRecalculateMetrics();
```
batwomankt, line 1333: I found that in case of exact length line breaks, there could be whitespace that remains untrimmed - the addition of this line here would take care of that (without losing the benefits of the fix added for issue 367)

tests/SixLabors.Fonts.Tests/Issues/Issues_367.cs
batwomankt: Yes, how embarrassing...I didn't push my last commit before the end of day

JimBobSquarePants: Happens to the best of us 😄

JimBobSquarePants (approved):
This all looks great. Thanks for the contribution!

batwomankt:
> This all looks great. Thanks for the contribution!
> 
> EDIT.
> 
> Looks like the tests still fail?

Yeah, I had missed one final spot where both CDK with break all and non-CDK needed the extra trimming!

PR #405, opened 2024-06-28
Apply kerning table offset to correct glyph.

JimBobSquarePants:
### Prerequisites

### Description

Updates the library to apply the kerning offset provided by the KERN table against the correct glyph. 

Note. This is a breaking change. The semantics for several public methods were incorrect and required updating.


tocsoft (approved):
Looks good

I doubt the breaking change would really impact anyone anyway and that particular API might not even be fully fit for purpose in all fonts either as its only giving the kern table kernings and not the gpos ones (which theoretically should be the same, but there nothing stopping them being different)

JimBobSquarePants:
@tocsoft Thanks for the quick review! I agree regarding the API. I remember someone asking for it, but I don't think it's all that useful.

PR #409, opened 2024-07-29
Fix memory leak in VerticalMetricsTable

Bykiev:
This PR fixes memory leak in `VerticalMetricsTable`

### DIFF
--- a/src/SixLabors.Fonts/Tables/General/VerticalMetricsTable.cs
+++ b/src/SixLabors.Fonts/Tables/General/VerticalMetricsTable.cs
@@ -47,7 +47,10 @@ internal sealed class VerticalMetricsTable : Table
             return null;
         }
 
-        return Load(binaryReader, headTable.NumberOfVMetrics, profileTable.GlyphCount);
+        using (binaryReader)
+        {
+            return Load(binaryReader, headTable.NumberOfVMetrics, profileTable.GlyphCount);
+        }
     }
 
     public static VerticalMetricsTable Load(BigEndianBinaryReader reader, int metricCount, int glyphCount)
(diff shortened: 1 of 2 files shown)


src/SixLabors.Fonts/Tables/General/VerticalMetricsTable.cs
```diff
@@ -47,7 +47,10 @@ internal short GetTopSideBearing(int glyphIndex)
             return null;
         }
 
-        return Load(binaryReader, headTable.NumberOfVMetrics, profileTable.GlyphCount);
+        using (binaryReader)
```
JimBobSquarePants, line 50: Nice job finding this! Did you find it profiling or just by chance?

JimBobSquarePants (approved):
Thanks for this. Much appreciated!

src/SixLabors.Fonts/Tables/General/VerticalMetricsTable.cs
Bykiev: I've used [Roslyn profiler](https://github.com/DotNetAnalyzers/IDisposableAnalyzers), there are some other issues, but due to used patterns in Fonts/ImageSharp I still investigating if it can be fixed (there are some non-sealed classes with non-virtual Dispose and some classes has a local variables, where class is not an owner of it). For. ex., `FontReader` class has a variable `stream`, but it has no Dispose, but some other classes implements it

JimBobSquarePants: I'd be careful with that analyzer, I've seen many, many false positives. 

`FontReader` doesn't own the input stream so shouldn't be responsible for its lifecycle. It's perfectly fine for the type to use a private variable to hold the reference.

Bykiev: Yes, it'll should be used carefully, because of such FPs

Bykiev: I'll prepare some interesting cases and will add an appropriate discussion if needed, still need some time to research all warning...

PR #411, opened 2024-08-01
Fix IDisposable usage

Bykiev:
This PR fixes a memory leak in `FontReader` class with `Woff2` fonts. Because of this `FontReader` class now implements `IDisposable` interface. 
This PR also wraps disposable types with using statements to remove warnings.

### DIFF
--- a/src/SixLabors.Fonts/Buffer{T}.cs
+++ b/src/SixLabors.Fonts/Buffer{T}.cs
@@ -26,7 +26,7 @@ internal ref struct Buffer<T>
         this.buffer = ArrayPool<byte>.Shared.Rent(bufferSizeInBytes);
         this.length = length;
 
-        ByteMemoryManager<T> manager = new(this.buffer);
+        using ByteMemoryManager<T> manager = new(this.buffer);
         this.Memory = manager.Memory.Slice(0, this.length);
         this.span = this.Memory.Span;
 
--- a/src/SixLabors.Fonts/FontReader.cs
+++ b/src/SixLabors.Fonts/FontReader.cs
@@ -9,21 +9,21 @@ using SixLabors.Fonts.Tables.Woff;
 
 namespace SixLabors.Fonts;
 
-internal sealed class FontReader
+internal sealed class FontReader : IDisposable
 {
     private readonly Stream stream;
     private readonly Dictionary<Type, Table> loadedTables = new();
     private readonly TableLoader loader;
+    private readonly bool disposable;
 
     internal FontReader(Stream stream, TableLoader loader)
     {
         this.loader = loader;
 
         Func<BigEndianBinaryReader, TableHeader> loadHeader = TableHeader.Read;
-        long startOfFilePosition = stream.Position;
 
         this.stream = stream;
-        var reader = new BigEndianBinaryReader(stream, true);
+        using var reader = new BigEndianBinaryReader(stream, true);
 
         // we should immediately read the table header to learn which tables we have and what order they are in
         uint version = reader.ReadUInt32();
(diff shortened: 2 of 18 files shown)


src/SixLabors.Fonts/Buffer{T}.cs
```diff
@@ -26,7 +26,7 @@ public Buffer(int length)
         this.buffer = ArrayPool<byte>.Shared.Rent(bufferSizeInBytes);
         this.length = length;
 
-        ByteMemoryManager<T> manager = new(this.buffer);
+        using ByteMemoryManager<T> manager = new(this.buffer);
```
JimBobSquarePants, line 29: `Dispose()` doesn't actually do anything for this type but given that is an implementation detail disposing is probably appropriate. 

However, the manager should be held as a private field of the parent `Buffer<T>` and disposed at the end of the parent lifecycle.

src/SixLabors.Fonts/FontReader.cs
```diff
@@ -85,13 +85,14 @@ internal FontReader(Stream stream, TableLoader loader)
...
 
+            this.disposable = true;
+
             byte[] compressedBuffer = reader.ReadBytes((int)totalCompressedSize);
             var decompressedStream = new MemoryStream();
             using var input = new MemoryStream(compressedBuffer);
             using var decompressor = new BrotliStream(input, CompressionMode.Decompress);
             decompressor.CopyTo(decompressedStream);
             decompressedStream.Position = 0;
-            this.stream.Dispose();
```
JimBobSquarePants, line 94: Yeah, that's a mistake. The parent should never dispose of a field it doesn't own.

src/SixLabors.Fonts/FontReader.cs
```diff
@@ -197,4 +198,12 @@ public bool TryGetReaderAtTablePosition(string tableName, [NotNullWhen(returnVal
         reader = header?.CreateReader(this.stream);
         return reader != null;
     }
+
+    public void Dispose()
+    {
+        if (this.disposable)
```
JimBobSquarePants, line 204: You need two `bool` fields.

1. `isDisposed` to prevent multiple calls.
2. `ownedStream` to specify ownership of the stream.

src/SixLabors.Fonts/Tables/Cff/CffDataDicEntry.cs
```diff
@@ -26,13 +26,13 @@ public override string ToString()
         {
             if (i > 0)
             {
-                builder.Append(" ");
+                builder.Append(' ');
```
JimBobSquarePants, line 29: This should be slightly more performant but please try to keep changes relevant.

JimBobSquarePants (commented):
Thanks for looking into this!

The changes look good just with some minor updates required. Looking at this PR reminds me I need to do a thorough style cleanup of the codebase!

src/SixLabors.Fonts/Tables/Cff/CffDataDicEntry.cs
Bykiev: I'll rollback this change, thanks!

src/SixLabors.Fonts/Buffer{T}.cs
Bykiev: Sorry, didn't get the idea about the purpose of private field. You can contribute to my repo if you wish.

src/SixLabors.Fonts/Tables/Cff/CffDataDicEntry.cs
JimBobSquarePants: Oh, no sorry, I don't need you to do that. It's a good change, just surprising 😄

JimBobSquarePants (approved):
Many thanks for this!

Bykiev:
> Many thanks for this!

Thank you for your hard work! I'll continue with disposable in `ImageSharp` project, it seems to be more complicated...

PR #424, opened 2024-12-03
Try Improve GC

Cn-mjt44:
# Improvment:
 1. Use ArrayBuffer to create array
 2. Use struct object to make object unmanaged


src/SixLabors.Fonts/GlyphPositioningCollection.cs
```diff
@@ -17,16 +18,18 @@ internal sealed class GlyphPositioningCollection : IGlyphShapingCollection
...
-    private readonly List<GlyphPositioningData> glyphs = new();
+    private GlyphPositioningData[] glyphs = ArrayPool<GlyphPositioningData>.Shared.Rent(16);
 
     /// <summary>
     /// Initializes a new instance of the <see cref="GlyphPositioningCollection"/> class.
     /// </summary>
     /// <param name="textOptions">The text options.</param>
     public GlyphPositioningCollection(TextOptions textOptions) => this.TextOptions = textOptions;
 
+    ~GlyphPositioningCollection() => ArrayPool<GlyphPositioningData>.Shared.Return(this.glyphs);
```
JimBobSquarePants, line 29: In C#, relying on a destructor (finalizer) without implementing `IDisposable` is typically discouraged and rarely sufficient. 

- The finalizer (`~ClassName`) runs during garbage collection.
- The timing of garbage collection is non-deterministic, so you cannot guarantee when (or even if) the finalizer will execute.
- Adding a finalizer to a class can degrade performance because the object is placed in a finalization queue, requiring at least two garbage collection cycles to be reclaimed.

src/SixLabors.Fonts/ArrayBuilder{T}.cs
```diff
@@ -153,23 +154,111 @@ private void EnsureCapacity(int min)
...
 
             this.data = array;
         }
     }
 
+    public readonly bool Contains(T item)
+    {
+        for (int i = 0; i < this.size; i++)
+        {
+            if (this[i].Equals(item))
```
JimBobSquarePants, line 172: This introduces boxing which incurs allocating memory on the heap.

src/SixLabors.Fonts/ArrayBuilder{T}.cs
```diff
@@ -153,23 +154,111 @@ private void EnsureCapacity(int min)
...
+        {
+            throw new ArgumentOutOfRangeException(nameof(count));
+        }
+
+        this.Length -= count;
+    }
+
+    public readonly void Sort() => this.data!.AsSpan(0, this.size).Sort();
+
+    public unsafe void Free()
```
JimBobSquarePants, line 246: Why Explicit `Free()` Methods Are a Bad Idea

1. **Breaks .NET Standards**: The `IDisposable` pattern is the established and widely understood approach for resource management in .NET. Introducing a custom `Free()` method deviates from this standard, making the code less intuitive for others.

2. **No `using` Support**: `IDisposable` integrates with the `using` statement, ensuring deterministic cleanup. A `Free()` method requires manual calls, increasing the risk of resource leaks due to missed invocations.

3. **No Safety Net**: Without a fallback like a finalizer, forgotten `Free()` calls can lead to unmanaged resource leaks, especially in edge cases like exceptions or improper usage.

4. **Inconsistent API Design**: Using `Free()` makes the API harder to use and understand, as developers expect `Dispose()` for cleanup tasks.

I would replace `Free()` with the `IDisposable` pattern in types where we can to ensure adherence to .NET conventions, improve usability, and reduce the potential for resource mismanagement. I don't think this type is one of them.

src/SixLabors.Fonts/Unicode/BidiData.cs
```diff
@@ -172,4 +174,15 @@ public ArraySlice<sbyte> GetTempLevelBuffer(int length)
         this.tempLevelBuffer.Clear();
         return this.tempLevelBuffer.Add(length, false);
     }
+
+    public void Free()
```
JimBobSquarePants, line 178: From C# 8 onwards ref structs can add Dispose methods which mean the compiler will allow `using` statements.

JimBobSquarePants (changes requested):
I appreciate your effort, however, I must decline this PR. 

The modifications present numerous deviations from standard .NET patterns, and the introduced code duplicates the functionality of existing .NET types without accompanying tests, leading to potential maintenance challenges.

Ideally, memory profiling should be conducted (JetBrains Rider offers this capability with unit tests) to pinpoint the primary issues. Subsequently, we can focus on those specific areas for improvement.

src/SixLabors.Fonts/ArrayBuilder{T}.cs
Cn-mjt44: > Why Explicit `Free()` Methods Are a Bad Idea
> 
> 1. **Breaks .NET Standards**: The `IDisposable` pattern is the established and widely understood approach for resource management in .NET. Introducing a custom `Free()` method deviates from this standard, making the code less intuitive for others.
> 2. **No `using` Support**: `IDisposable` integrates with the `using` statement, ensuring deterministic cleanup. A `Free()` method requires manual calls, increasing the risk of resource leaks due to missed invocations.
> 3. **No Safety Net**: Without a fallback like a finalizer, forgotten `Free()` calls can lead to unmanaged resource leaks, especially in edge cases like exceptions or improper usage.
> 4. **Inconsistent API Design**: Using `Free()` makes the API harder to use and understand, as developers expect `Dispose()` for cleanup tasks.
> 
> I would replace `Free()` with the `IDisposable` pattern in types where we can to ensure adherence to .NET conventions, improve usability, and reduce the potential for resource mismanagement. I don't think this type is one of them.

The reason here:
1. It has complex reference so I can't transfer it into struct.
2. this object is little object, so it will collect by GC gen 0. That is way more faster than GC gen 1 for large object. 
3. It only create 1 times when generate glyph layout.

src/SixLabors.Fonts/Unicode/BidiData.cs
Cn-mjt44: But ref struct can't import interface. Is that means I can announce a method called Dispose, so I can use using statements?

JimBobSquarePants: Yes, they don't implement the interface but the compiler spots the pattern and allows you to use it.

src/SixLabors.Fonts/ArrayBuilder{T}.cs
Cn-mjt44: I have copy the function from `List<T>.Contains` for this part. Should I finish this part?

JimBobSquarePants: Using memory pooling for this type is not a good idea. 

It’s problematic because it allows unrestricted types, leading to inefficient pooling and increased risk of pool exhaustion. Each type parameter combination creates a unique pool, fragmenting the memory and reducing reuse. If frequently requested types deplete their pools, allocations bypass pooling, negating performance benefits and increasing GC pressure.

That’s why the type does not already exist in the runtime and why our memory allocators have explicit constraints.

Cn-mjt44: The reason I do that:
1. ArrayPool will return the larger array inside or create new array for that.
2. ArrayPool will delete the array object when something rent it from pool, and add it back when return it.
And also, this type(struct) is internal type. So where it used and what the unrestricted types used are also knowned.
Everything in there are manual assign.

src/SixLabors.Fonts/ArrayBuilder{T}.cs
Cn-mjt44: I have replaced the `Free` to `Dispose`, this part should close.

PR #426, opened 2024-12-05
Remove all instances of HasFlag

JimBobSquarePants:
### Prerequisites

### Description

See https://github.com/SixLabors/Fonts/discussions/425#discussioncomment-11468049

>HasFlags sometimes still boxes (in tier 0 compilation) if JIT does not optimize it away.

@nolife99 I profiled a couple of unit tests and saw a reduction however it would be useful if you could give this a try by referencing the code directly and report back.

### DIFF
--- a/src/SixLabors.Fonts/Font.cs
+++ b/src/SixLabors.Fonts/Font.cs
@@ -290,7 +290,7 @@ public sealed class Font
             return metrics;
         }
 
-        if (this.RequestedStyle.HasFlag(FontStyle.Italic))
+        if ((this.RequestedStyle & FontStyle.Italic) == FontStyle.Italic)
         {
             // Can't find style requested and they want one that's at least partial italic.
             // Try the regular italic.
@@ -300,7 +300,7 @@ public sealed class Font
             }
         }
 
-        if (this.RequestedStyle.HasFlag(FontStyle.Bold))
+        if ((this.RequestedStyle & FontStyle.Bold) == FontStyle.Bold)
         {
             // Can't find style requested and they want one that's at least partial bold.
             // Try the regular bold.
--- a/src/SixLabors.Fonts/FontDescription.cs
+++ b/src/SixLabors.Fonts/FontDescription.cs
@@ -172,24 +172,24 @@ public class FontDescription
 
         if (os2 != null)
         {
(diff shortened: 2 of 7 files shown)


nolife99:
I did implement this myself yesterday (exactly the same as you did), and it showed a pretty good improvement in memory usage and GC performance

JimBobSquarePants:
@nolife99  You wouldn't have any screenshots you can share or some sample code I can use to reproduce would you?

nolife99:
```
    public Image<Rgba32> CreateBitmap(string text,
        string fontName,
        float fontSize,
        Vector2 padding,
        out Vector2 textureSize)
    {
        if (string.IsNullOrEmpty(text)) text = " ";

        const float dpi = 72f; 

        // replace below with a Font instance, '96 * fontSize / dpi' is the font size
        var font = getFont(fontName, 96 * fontSize / dpi, FontStyle.Regular);

        var foundMetrics = font.Family.TryGetMetrics(FontStyle.Regular, out var metrics);
        RichTextOptions options = new(font)
        {
            HorizontalAlignment = HorizontalAlignment.Center,
            VerticalAlignment = VerticalAlignment.Center,
            LineSpacing = foundMetrics ? Math.Max(metrics.VerticalMetrics.LineHeight * .001f, 1) : 1,
            Dpi = dpi
        };

        var measuredSize = TextMeasurer.MeasureAdvance(text, options);
        var width = (int)(measuredSize.Width + padding.X * 2 + 1);
        var height = (int)(measuredSize.Height + padding.Y * 2 + 1);

        textureSize = new(width, height);

        Image<Rgba32> bitmap = new(width, height);
        bitmap.Mutate(b =>
        {
            RichTextOptions drawOptions = new(font) { Origin = padding };
            b.DrawText(new(drawOptions) { Origin = padding + Vector2.One }, text, Color.White).DrawText(drawOptions, text, Color.FromRgba(0, 0, 0, 220));
        });

        return bitmap;
    }

    Image<Rgba32> generateGlyph(char c)
    {
        return CreateBitmap(c.ToString(), name, size, default, out measuredSize);
    }
```
This is a snippet of code from one of my projects, basically generateGlyph() is called for each character in a string
Sorry if there are any errors, I extracted this in a rush

Issue #434, opened 2024-12-16
One extra newline is not added when 2 explicit newlines are specified


jez9999:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of Fonts
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/Fonts/issues) and [closed](https://github.com/SixLabors/Fonts/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Fonts version

2.0.7

### Other Six Labors packages and versions

SixLabors.ImageSharp 3.1.6, SixLabors.ImageSharp.Drawing 2.1.4

### Environment (Operating system, version and so on)

Windows 10

### .NET Framework version

.NET 9

### Description

After fix #433, another issue has been created; now, a second explicit newline is not added when it should be.  See repro program below.

### Steps to Reproduce

```
internal class Program {
    static void Main(string[] args) {
        Console.WriteLine("Rendering test images...");

        var fontManager = new FontManager();

        Image imgTest3 = new Image<Rgba32>(550, 350, Color.White);
        imgTest3.Mutate(x =>
            x.DrawText(
                new RichTextOptions(fontManager.GetFont("Arial", 60, FontStyle.Regular)) {
                    Origin = new Vector2(50, 20),
                    WrappingLength = 400,
                },
                "- Lorem ipsullll\n\ndolor sit amet\n-consectetur elit",
                Color.Black
            )
        );
        imgTest3.SaveAsPng("c:/development/diff/testImg3.png");
        // ^ incorrect, no extra newline after: - Lorem ipsullll

        Image imgTest4 = new Image<Rgba32>(550, 350, Color.White);
        imgTest4.Mutate(x =>
            x.DrawText(
                new RichTextOptions(fontManager.GetFont("Arial", 60, FontStyle.Regular)) {
                    Origin = new Vector2(50, 20),
                    WrappingLength = 400,
                },
                "- Lorem ipsullll\n\n\ndolor sit amet\n-consectetur elit",
                Color.Black
            )
        );
        imgTest4.SaveAsPng("c:/development/diff/testImg4.png");
        // ^ incorrect, one rather than two extra newlines after: - Lorem ipsullll

        Console.WriteLine("Done.");
    }
}
```

### Images

[...]

Issue #431, opened 2024-12-13
Unnecessary text wrap when newline separates words instead of space


jez9999:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of Fonts
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/Fonts/issues) and [closed](https://github.com/SixLabors/Fonts/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Fonts version

2.0.6

### Other Six Labors packages and versions

SixLabors.ImageSharp 3.1.6, SixLabors.ImageSharp.Drawing 2.1.4

### Environment (Operating system, version and so on)

Windows 10

### .NET Framework version

.NET 9

### Description

When using `DrawText`, text is unnecessarily given a newline through text wrapping (ie. nothing is then rendered on the new line) when a newline character is used to separate two words rather than a space.

### Steps to Reproduce

```
internal class Program {
    static void Main(string[] args) {
        Console.WriteLine("Rendering test images...");

        var fontManager = new FontManager();

        Image imgTest1 = new Image<Rgba32>(550, 350, Color.White);
        imgTest1.Mutate(x =>
            x.DrawText(
                new RichTextOptions(fontManager.GetFont("Arial", 60, FontStyle.Regular)) {
                    Origin = new Vector2(50, 20),
                    WrappingLength = 400,
                },
                "- Lorem ipsullll dolor sit amet\n-consectetur elit",
                Color.Black
            )
        );
        imgTest1.SaveAsPng("c:/development/diff/testImg1.png");
        // ^ correct

        Image imgTest2 = new Image<Rgba32>(550, 350, Color.White);
        imgTest2.Mutate(x =>
            x.DrawText(
                new RichTextOptions(fontManager.GetFont("Arial", 60, FontStyle.Regular)) {
                    Origin = new Vector2(50, 20),
                    WrappingLength = 400,
                },
                "- Lorem ipsulll\ndolor sit amet\n-consectetur elit",
                Color.Black
            )
        );
        imgTest2.SaveAsPng("c:/development/diff/testImg2.png");
        // ^ correct, no wrapping yet

        Image imgTest3 = new Image<Rgba32>(550, 350, Color.White);
        imgTest3.Mutate(x =>
            x.DrawText(
[...]

PR #438, opened 2025-01-07
Rewrite Line Breaker.

JimBobSquarePants:
### Prerequisites

### Description

A complete rewrite of the line breaker to fix a series of issues including #434 I've not benchmarked this but in addition to fixing bugs the code is now doing far less work per code point so should be faster. 

I've also added testing that will generate output using ImageSharp.Drawing which allows us to better visualize the layout. Adding this helped identify multiple issues. The generation code is togglable via a compiler condition introduced via the `csproj` file. This allows us to turn them off should we introduce breaking changes. 

All output has been sense checked and compared to browser output to ensure correctness.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -155,7 +155,9 @@ jobs:
         if: failure()
         with:
           name: actual_output_${{ runner.os }}_${{ matrix.options.framework }}${{ matrix.options.runtime }}.zip
-          path: tests/Images/ActualOutput/
+          path: |
+              tests/Images/ActualOutput/
+              msbuild.log
 
       - name: Codecov Update
         uses: codecov/codecov-action@v4
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -157,7 +157,7 @@ jobs:
           name: actual_output_${{ runner.os }}_${{ matrix.options.framework }}${{ matrix.options.runtime }}.zip
           path: |
               tests/Images/ActualOutput/
-              msbuild.log
+              **/msbuild.binlog
 
       - name: Codecov Update
         uses: codecov/codecov-action@v4
--- a/.gitignore
+++ b/.gitignore
@@ -254,6 +254,8 @@ paket-files/
 *.sln.iml
 /samples/DrawWithImageSharp/Output
 
+# Tests
+**/Images/ActualOutput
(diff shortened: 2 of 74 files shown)


tocsoft (approved):
lgtm 👍

jez9999:
I've tried 2.0.8 against my bugs #431 and #434 and it seems to have fixed both issues. 👍 

I would note however that it appears to have changed what ImageSharp considers to be the bounding box of the rendered text, based on `WrappingLength`.  The first `imgTest1` output from #431 now wraps after "Lorem ipsullll" unless I slightly increase the `WrappingLength`.  Maybe the original behaviour was a bug that's also been fixed.

JimBobSquarePants:
> I've tried 2.0.8 against my bugs #431 and #434 and it seems to have fixed both issues. 👍
> 
> I would note however that it appears to have changed what ImageSharp considers to be the bounding box of the rendered text, based on `WrappingLength`. The first `imgTest1` output from #431 now wraps after "Lorem ipsullll" unless I slightly increase the `WrappingLength`. Maybe the original behaviour was a bug that's also been fixed.

I'm not sure I understand what you are saying. In your original issue the text wrapped following "Lorem ipsullll"

testImg1

jez9999:
With 2.0.6 I got:

testImg1

With 2.0.8 I get:

testImg1

jez9999:
Hmm, well I can't easily use `TextLayoutTestUtilities` because it's internal.  Can't it just be made public?

jez9999:
Also I noticed your test string is `- Lorem ipsullll\ndolor sit amet\n-consectetur elit`, instead try `- Lorem ipsullll dolor sit amet\n-consectetur elit` without the explicit newline and it wraps.

JimBobSquarePants:
> Hmm, well I can't easily use `TextLayoutTestUtilities` because it's internal. Can't it just be made public?

Why would I make a test utility that doesn’t get published with any code public? You know the code is right here yeah? You can simply clone the repo and run it? 

Please stop changing the goal posts and drip feeding me examples. I’m testing the strings provided in your issues and each test takes a lot of my time of which I have a limited capacity. 

This is not only not helpful but incredibly frustrating.

jez9999:
OK well I'm sorry, I didn't intend that.

I tested it with the string `- Lorem ipsullll dolor sit amet\n-consectetur elit` and it appears to incorrectly line wrap the "ipsullll" in 2.0.8:

testImg1

PR #447, opened 2025-01-30
Fix Tracking of Last Line Break

JimBobSquarePants:
### Prerequisites

### Description

Fixes #466

I had a bug in the tracking of the last wrapped position, have fixed that, cleaned up a little and added some caching so that the wrap measurements are not repeated unnecessarily. 

@jez9999 if you have the time and fancy running your code against this branch to see if anything else appears that would be very useful. You seem to have the magic touch for finding issues.

### DIFF
--- a/src/SixLabors.Fonts/TextLayout.cs
+++ b/src/SixLabors.Fonts/TextLayout.cs
@@ -1327,8 +1327,6 @@ internal static class TextLayout
 
         public int Count => this.data.Count;
 
-        public int LeadCodePointIndex { get; private set; } = -1;
-
         public float ScaledLineAdvance { get; private set; }
 
         public float ScaledMaxLineHeight { get; private set; } = -1;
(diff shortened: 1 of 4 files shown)


jez9999:
Is there a way for me to run just this repo with the other ImageSharp NuGet packages or do I need to remove all NuGet package references and clone all the repos?

JimBobSquarePants:
> Is there a way for me to run just this repo with the other ImageSharp NuGet packages or do I need to remove all NuGet package references and clone all the repos?

You should be able to use the NuGet packages and add Fonts as a project reference pointing to this branch. That’s how we do the sample project in this repo.

jez9999:
`.Drawing` seems to have a dependency of Fonts >= 2.0.8 so I'm not sure I can just remove fonts without that too?

JimBobSquarePants:
It looks like it's not finding your standard NuGet package source. I wonder if it's a submodule thing, do you have a populated submodule `shared-infrastructure` in your Fonts clone?

jez9999:
Ah.  Darn git behaviour, should init all submodules by default I say.

Still now getting various errors along the lines of this for some reason: Duplicate 'System.Reflection.AssemblyCompanyAttribute' attribute

jez9999:
OK well I got it building now, but as I suspected, it seems to be using the old Fonts 2.0.8 pulled in by the Drawing NuGet package rather than the referenced Fonts project.

jez9999:
Yep, it seems to be good now with the stuff I have.

tocsoft (approved):

PR #342, opened 2023-07-14
WIP: Compact Font Format (CFF) Version 2

brianpopow:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/Fonts/pulls) open
- [x] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [ ] I have provided test coverage for my change (where applicable)

### Description

This PR aims to add support for CFF2 format. Its still a work in progress.

The Specification can be found [here](https://learn.microsoft.com/en-us/typography/opentype/otspec181/cff2).
Relevant tables for the font variations are:
- [fvar](https://learn.microsoft.com/de-de/typography/opentype/spec/fvar)
- [gvar](https://learn.microsoft.com/de-de/typography/opentype/spec/gvar)
- [avar](https://learn.microsoft.com/de-de/typography/opentype/spec/avar)
- [hvar](https://learn.microsoft.com/de-de/typography/opentype/spec/hvar)

Example font for CFF2: 
[RobotoFlex.zip](https://github.com/SixLabors/Fonts/files/12050890/RobotoFlex.zip)

TODO:
- The parsing of the ItemVariations data does not match what fontkit parses, see [comment](https://github.com/SixLabors/Fonts/issues/331#issuecomment-1632229980)
- We need some method to create a new font based on the variation settings a user chooses. For example: `font.GetVariation(variationOptions)`. The equivalent in fontkit would be `font.getVariation({wght: 300})`

PR #465, opened 2025-10-28
Fixed errors

stefannikolei:
I opened your PR in Rider and got many errors

Fixed those:
- Target type new
- Collection Expressions
- Usage of var


stefannikolei:
@JimBobSquarePants Was VS silent about those things?

Perhaps adding  `<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>` to Directory.Build.props will make VS to show the errors too?

JimBobSquarePants:
That’s interesting… There were warnings but nothing blocking. I didn’t want to change things I didn’t need to yet to avoid conflicts with #342

stefannikolei:
Those errors are not blocking in rider. 
But everything is just red 🫣

PR #540, opened 2026-07-03
Fix TrueTypeInterpreter pool state pollution

christianrondeau:
### Prerequisites

### Description

Some of our tests were non-deterministic. Fonts drawing would _sometimes_ be different when running under load (here's a repro of the behavior we are seeing: [truetype_pool_leak.zip](https://github.com/user-attachments/files/29643952/truetype_pool_leak.zip))

It turns out to be caused by the TrueTypeInterpreter pool not fully reinitializing between uses, which has unexpected side effects when rendering different font sizes. Related to https://github.com/SixLabors/Fonts/pull/486.

A simpler repro is in the commit's unit tests. Note that I added the state reset in SetControlValueTable because that's where the other Reset\Clear were, but it might be better to put this reset code when pulling an instance out of the pool instead (Reset method called after this.interpreterPool.Get()), could be a refactor if you want. Also note that I'm resetting a bit more than strictly necessary to avoid other potential issues I might have missed.


JimBobSquarePants:
Closed, replaced by #541

PR #542, opened 2026-07-10
Faux italic when the font does not support it

christianrondeau:
### Prerequisites

### Description

Some fonts may not provide native italic, and most text rendering engines will fallback to faux-italic (simply slant the text). ImageSharp does not provide that function, which means italic text is simply rendered upright. The italic slant specifications are vague, but the default seems to be 14 degrees (https://www.w3.org/TR/css-fonts-4/).

Disclaimer: this code was written using AI (Copilot with Claude Opus 4.8), and reviewed by me.

image

This solves the italic part of https://github.com/SixLabors/ImageSharp.Drawing/discussions/408 - the weight (bold) may be a bit more complex and will be provided as a separate PR.

### DIFF
--- a/src/SixLabors.Fonts/FontGlyphMetrics.cs
+++ b/src/SixLabors.Fonts/FontGlyphMetrics.cs
@@ -16,6 +16,13 @@ public abstract class FontGlyphMetrics
 {
     private static readonly Vector2 YInverter = new(1, -1);
 
+    /// <summary>
+    /// The horizontal shear applied to synthesize an oblique (faux italic) slant when an italic
+    /// style is requested but the resolved font face provides no italic face. This equals the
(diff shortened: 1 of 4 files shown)


JimBobSquarePants (approved):
Thanks for this. I’m going to accept this PR with a caveat.
The approach is directionally correct: synthetic italic belongs in Fonts rather than in individual renderers, and the chosen skew is close to browser/Skia behavior. Chromium/Skia use a `1/4` skew for synthetic italic, which is approximately a 14 degree slant, and synthetic italic should not change advance width.
I’ll likely need to adjust this after merging into my local branch because the rendering path has changed there, but the behavior this PR is aiming for is the behavior I want.

christianrondeau:
Thank you, looking forward to seeing the new rendering branch (ImageSharp has among the best code I've seen, so I'm always keen to learn from it!)

PR #543, opened 2026-07-10
Faux bold when the font does not support it

christianrondeau:
### Prerequisites

### Description

A follow-up to https://github.com/SixLabors/Fonts/pull/542 - should be merged after the italic one.

Bold is much more complex than italic, because it requires dilating glyphs, although it does not require changing the advance (I thought it would require it but comparing to Chrome, it looks fine with just widening the glyphs).

Disclaimer: this code was written using AI (Copilot with Claude Opus 4.8), and reviewed by me. However, while the italic PR is fairly simple to understand, I don't have the knowledge to personally fully verify `EmboldeningGlyphRenderer`. This being said, the logic looks sound and renders correctly.

image

Thanks for your consideration!

### DIFF
--- a/src/SixLabors.Fonts/FontGlyphMetrics.cs
+++ b/src/SixLabors.Fonts/FontGlyphMetrics.cs
@@ -23,6 +23,14 @@ public abstract class FontGlyphMetrics
     /// </summary>
     private const float SyntheticObliqueSkew = 0.24932839F;
 
+    /// <summary>
+    /// The outward outline offset, expressed as a fraction of the em size, applied to synthesize
+    /// a bold (faux bold) weight when a bold style is requested but the resolved font face provides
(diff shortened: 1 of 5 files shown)


JimBobSquarePants (approved):
Thanks for this. I’m going to accept this PR as an initial integration point, but with some important caveats.

Synthetic bold should be owned by Fonts so that all renderers receive the same glyph geometry. However, the current emboldening implementation should not be treated as final. I want to follow browser layout behavior, meaning synthetic bold should not change layout advance, but I also want the actual outline dilation logic to be much closer to FreeType’s outline emboldening algorithm.

So I’m going to merge this to unblock the feature, then follow up by replacing/refining the dilation algorithm and adding stronger tests around counters, mixed contour winding, CFF/TrueType outlines, and bounds behavior.

christianrondeau:
Thank you; let me know if you'd like me to spend more time on any of it!

PR #556, opened 2026-08-28
Fix signed fractional CFF coordinates

Artromskiy:
## Problem

CFF Type 2 32-bit numbers are signed 16.16 fixed-point values. In `v3.1.0`, the fractional word was read as a signed `short`. Whenever bit 15 of the fractional word was set, the decoded value was one design unit too small. On a CFF glyph this error accumulated across relative curve coordinates and produced visibly incorrect outlines.

The issue was reproduced with glyph 54 (`R`) from an external Igrunok OTF. The font is not included because its local license does not establish that redistribution in this test repository is permitted. The regression uses the raw Type 2 prefix from that glyph, without distributing the font.

## Fix

Read the integer word as signed and the fractional word as unsigned, which is the correct signed 16.16 decomposition.

## Tests

- `dotnet build src/SixLabors.Fonts/SixLabors.Fonts.csproj -c Release` — passed, 0 warnings/errors.
- Targeted `CffEvaluationEngineTests` — 2 passed.
- External glyph 54 direct renderer probe: 25 commands and 65 points match CoreText/Skia after the expected Y reflection; maximum coordinate difference was 0.00005 px at 280 px/em.

The full test suite was not used as acceptance for this local clone because test fonts were unavailable as LFS objects after a smudge/download timeout.

### DIFF
--- a/src/SixLabors.Fonts/Tables/Cff/SimpleBinaryReader.cs
+++ b/src/SixLabors.Fonts/Tables/Cff/SimpleBinaryReader.cs
@@ -78,9 +78,9 @@ internal ref struct SimpleBinaryReader
         byte b0 = this.buffer[this.Position];
         this.Position += 4;
 
-        // This number is interpreted as a Fixed; that is, a signed number with 16 bits of fraction
+        // The integer word is signed, while the fractional word is an unsigned 16-bit value.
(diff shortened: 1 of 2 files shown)


Artromskiy:
Closing this PR to recreate it with the corrected Artromskiy commit identity.

